PR #2, opened 2019-07-15
Add environment file for conda installation

jlenain:
This very simple environment file can be used to create an anaconda environment called `nectarchain-dev` with:

`conda env create -f environment.yml`

The problem at this stage is that `ctapipe` is installed directly with `conda` from the `cta-observatory` repository, i.e. using the last tagged version available, which is 0.6.2 as of today.

Installing an environment from git master by using instead:

```
# A conda environment with all necessary packages for nectarchain developers
name: nectarchain-dev
channels:
  - default
dependencies:
  - python=3.7  # nail the python version, so conda does not try upgrading / dowgrading
  - pip
  - pip:
      - https://github.com/cta-sst-1m/protozfitsreader/archive/v1.4.2.tar.gz
      - git+git://github.com/cta-observatory/ctapipe.git
      - git+git://github.com/cta-observatory/ctapipe_io_nectarcam.git
```
will work, but an attempt to read NectarCAM data:

```
from ctapipe.io import event_source, EventSeeker
path = 'obs/NectarCAM.Run1250.0000.fits.fz'
reader = event_source(input_url=path)
seeker = EventSeeker(reader)
currentevent = seeker[0]
```
fails with:
```
Seeking to event by looping through events... (potentially long process)
Traceback (most recent call last):
  File "/home/jlenain/local/src/python/anaconda/envs/nectarchain-dev/lib/python3.7/site-packages/ctapipe/io/eventseeker.py", line 182, in __getitem__
    event = self._reader._get_event_by_index(item)
AttributeError: 'NectarCAMEventSource' object has no attribute '_get_event_by_index'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jlenain/local/src/python/anaconda/envs/nectarchain-dev/lib/python3.7/site-packages/ctapipe/io/eventseeker.py", line 191, in __getitem__
    event = self._get_event_by_index(item)
  File "/home/jlenain/local/src/python/anaconda/envs/nectarchain-dev/lib/python3.7/site-packages/ctapipe/io/eventseeker.py", line 218, in _get_event_by_index
    for event in self._source:
  File "/home/jlenain/local/src/python/anaconda/envs/nectarchain-dev/lib/python3.7/site-packages/ctapipe/io/eventsource.py", line 217, in __iter__
    for event in self._generator():
[...]

PR #5, opened 2019-07-31
Setup environment

FrancaCassol:
I add here some environment files, that should also activate the travis check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,40 @@
+language: python
+
+matrix:
+    include:
+        python: 3.6
+        os: linux
+        env:
+            - CTAPIPE_VERSION=master
+            - PROTOZFITS_VERSION=v1.4.2
+            - CTAPIPE_IO_NECTARCAM_VERSION=master
+
+
+before_install:
+
+    # Use utf8 encoding. Should be default, but this is insurance
+    # against future changes
+    - export PYTHONIOENCODING=UTF8
+    - export MPLBACKEND=Agg
+
+    - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    - bash miniconda.sh -b -p $HOME/miniconda
+    - . $HOME/miniconda/etc/profile.d/conda.sh
+    - hash -r
+    - conda config --set always_yes yes --set changeps1 no
+    - conda update -q conda  # get latest conda version
+    - conda info -a # Useful for debugging any issues with conda
+
+install:
+    - conda env create --name cta --file environment.yml
+    - conda activate cta
+    - pip install https://github.com/cta-observatory/ctapipe/archive/$CTAPIPE_VERSION.tar.gz
+    - pip install https://github.com/cta-sst-1m/protozfitsreader/archive/$PROTOZFITS_VERSION.tar.gz
+    - pip install https://github.com/cta-observatory/ctapipe_io_nectarcam/archive/$CTAPIPE_IO_NECTARCAM_VERSION.tar.gz
+    - python setup.py install
+
+
+script:
+    - pytest
+
+
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,13 @@
+name: cta
+channels:
+  - default
+  - cta-observatory
+  - conda-forge
+dependencies:
+  - python>3.5
+  - ctapipe
+  - gammapy
+  - pip:
+      - pytest_runner
+      - pytest-ordering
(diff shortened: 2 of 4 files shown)


environment.yml
```diff
@@ -0,0 +1,13 @@
+name: cta
+channels:
+  - default
+  - cta-observatory
+  - conda-forge
+dependencies:
+  - python>3.5
+  - ctapipe
```
jlenain: Something I do not understand here: we agreed in #2 that we should depend on the master of ctapipe, while here the dependency on ctapipe is bound to its last release in conda, which can be behind the master.

environment.yml
```diff
@@ -0,0 +1,13 @@
...
+  - gammapy
+  - pip:
+      - pytest_runner
+      - pytest-ordering
+      - https://github.com/cta-sst-1m/protozfitsreader/archive/v1.4.2.tar.gz
```
jlenain: Here, as discussed in #2, I think we should also depend on `ctapipe-extra`, so that the geometry files of the different cameras are known.

.travis.yml
```diff
@@ -0,0 +1,40 @@
...
+    - . $HOME/miniconda/etc/profile.d/conda.sh
+    - hash -r
+    - conda config --set always_yes yes --set changeps1 no
+    - conda update -q conda  # get latest conda version
+    - conda info -a # Useful for debugging any issues with conda
+
+install:
+    - conda env create --name cta --file environment.yml
+    - conda activate cta
+    - pip install https://github.com/cta-observatory/ctapipe/archive/$CTAPIPE_VERSION.tar.gz
```
jlenain: This appears to be inconsistent with the `envrionment.yml` file proposed: the conda environment will first install `ctapipe` from conda, and then again from `pip` here.

environment.yml
FrancaCassol: Actually a developer has to take care to test his code with the master version. While a user will install the master version when downloading his code (as its is said in the README [https://github.com/FrancaCassol/nectarchain/tree/setup_environment](https://github.com/FrancaCassol/nectarchain/tree/setup_environment)

environment.yml
FrancaCassol: The dependence from ctapipe-extra is already set by the ctapipe installation

.travis.yml
FrancaCassol: I think that this installation is only for travis which will run the tests. Unfortunately I don't know how to test it before accepting the PR

environment.yml
jlenain: For a developer, this is clear. But I thought it would be easier for users to just stick with a conda environment, without the need to pip install additional codes. Hence my suggestion in #2 to remove the installation of `ctapipe` from the cta-observatory conda channel, but rather integrate the pip installation in the environment itself.

environment.yml
jlenain: This is related to my last reply above: if `ctapipe` is installed via `pip` in this conda environment file, then a dependency on `ctapipe-extra` should be added here, presumably installing via `pip` too. If we stick with the conda installation of `ctapipe`, then I agree with you. But in the later case, `ctapipe` would be installed twice: once via conda, and a second time through the pip installation. Isn't there a risk of conflicting versions in such a case ?

.travis.yml
jlenain: I think I get it: this is actually consistent with the instructions provided in the `README`, but only if we don't install `ctapipe` via `pip` via the environment file itself (see replies to comments above). If `ctapipe` is to be installed via `pip` in the environment file, then this statement in the `travis` configuration file is redundant with the instruction `conda env create --name cta --file environment.yml`.

jlenain:
If this PR is accepted, we should give up on #2, right ?

FrancaCassol:
> If this PR is accepted, we should give up on #2, right ?

Yes, I guess. This is the way done for the other packages, but if #2  works and you add also travis we can also accepts #2 (sorry, I understood there were problems with it). I will be on holiday for three weeks starting from tomorrow. I let you choose ;-) In the case we go for #2 I will update my PR for the calibration code.

.travis.yml
kosack: is there a reason you don't just do `conda install -c cta-observatory ctapipe=$CTAPIPE_VERSION` ?

FrancaCassol:
Hi @jlenain, 
could we converge with this PR or PR #2
I would like to get the calibration code accepted.

jlenain:
Hi @FrancaCassol ,

On August 29, by mail, we evoked the possibility to split into two branches:
- one fixed to ctapipe 0.7
- another branch evolving with master
In order to move on, for the time being, I propose to accept this PR and reject PR #2, agreed ?

PR #6, opened 2019-08-01
Calibration code

FrancaCassol:
I propose in this PR some tools and notebooks for performing a preliminary estimation of the calibration coefficients based on pedestal and flat-field events.
These tools use the Component classes of the flatfield and pedestal modules located in ctapipe/calib/camera. A more detail study could bring to create new classes specific to nectarCAM, which should be written in a directory nectarchain/calib/camera and eventually then transferred to ctapipe


jlenain:
Hi @FrancaCassol ,
Many thanks for the new push, everything is working like a charm. I am merging now.
Thanks !

PR #7, opened 2019-10-17
Updating protozfitsreader version to read native charges.

sizun:
Update to latest reader is necessary to avoid warnings when reading the latest data files and read native charges. See cta-sst-1m/protozfitsreader#62


sizun:
> just let me know when you're done
Hi @jlenain , I think that's it.

jlenain:
> Hi @jlenain , I think that's it.
OK, thanks a lot.

PR #11, opened 2022-02-20
Adding camera temperature monitoring to the DQM pipeline

hashkar:
in this pull request, the camera temperature monitoring module is added as a separate class. It find the corresponding SQL file and computes the camera's average temperature during the times of the run.

PR #14, opened 2022-04-20
Debugging the code after updates and adding camera monitoring module

hashkar:


jlenain:
Hi @hashkar !
Thanks a lot, but why a new PR ? Shouldn't we follow up in https://github.com/cta-observatory/nectarchain/pull/11 ?

hashkar:
Hi @jlenain ,
The problem is that I changed my Mac and had to download the software again. As there was lot of changes git didn't let me pull, push or merger. So I created a new git pull. This has the same changes as the last one with additional debugging.

jlenain:
Hi @hashkar ,
OK, I understand. I am running tests right now on a couple of runs, and I think we can quickly merge this.
Thanks again!

jlenain:
Hi @hashkar !
Many thanks again for this PR! I have run tests on a couple of runs. Everything went fine, I am merging this into `master`now.

PR #19, opened 2022-11-26
Update environment.yml and setup.py

mdpunch:
Have to pin numpy to 1.22, as @maxnoe remarks "ctapipe 0.12 depends on astropy 4, which is incompatible with the most recent numpy release 1.23"

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.8 # nail the python version, so conda does not try upgrading / downgrading
   - conda-forge::ctapipe=0.12
+  - numpy=1.22.4 # higher versions don't work due to astropy4 dependency in ctapipe0.12
   - jupyterlab
   - conda-forge::protozfits=2.0
   - pytest-runner
--- a/environment.yml
+++ b/environment.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - python=3.8 # nail the python version, so conda does not try upgrading / downgrading
   - conda-forge::ctapipe=0.12
-  - numpy=1.22.4 # higher versions don't work due to astropy4 dependency in ctapipe0.12
+  - numpy=1.22 # higher versions don't work due to astropy4 dependency in ctapipe0.12
   - jupyterlab
   - conda-forge::protozfits=2.0
   - pytest-runner
--- a/environment.yml
+++ b/environment.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - python=3.8 # nail the python version, so conda does not try upgrading / downgrading
   - conda-forge::ctapipe=0.12
-  - numpy=1.22 # higher versions don't work due to astropy4 dependency in ctapipe0.12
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.8 # nail the python version, so conda does not try upgrading / downgrading
   - conda-forge::ctapipe=0.12
+  - numpy=1.22.4 # higher versions don't work due to astropy4 dependency in ctapipe0.12
```
maxnoe, line 7: I'd suggest not pinning to `1.22.4` as that would prevent updating to a possible 1.22.5 bug-fix release

maxnoe:
Please also add it in `setup.py` (this is the actually important requirement, the env is just "nice-to-have".

environment.yml
mdpunch: Yes, thanks.  Trying to modify the commit now...

mdpunch:
Question for @jlenain , should these be in the `setup.py` of `nectarchain` also?
For the moment, there are no pinned versions there.

mdpunch:
i.e.,
```
        'astropy~=4.2',
        'ctapipe~=0.12',
        'numpy~=1.22.4',
```

maxnoe:
`nectarchain` should depend at least on `ctapipe_io_nectarcam`, right?

mdpunch:
> `nectarchain` should depend at least on `ctapipe_io_nectarcam`, right?

I would have thought, but it doesn't seem to!

The installation of `ctapipe_io_nectarcam`  is a separate `git clone` and `pip install`.

I guess there are reasons...

maxnoe:
The reason will be that `ctapipe_io_nectarcam` is not released to pypi, which should be done.

jlenain:
> Question for @jlenain , should these be in the `setup.py` of `nectarchain` also? For the moment, there are no pinned versions there.

Hi @mdpunch , yes, please, also add the `numpy` pinned version to `setup.py`. Thanks!

environment.yml
```diff
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.8 # nail the python version, so conda does not try upgrading / downgrading
   - conda-forge::ctapipe=0.12
+  - numpy~=1.22.4 # higher versions >=1.23 don't work due to astropy4 dependency in ctapipe0.12
```
jlenain, line 7: Hi @mdpunch ,

As noted already by @maxnoe , the syntax for the environment specifications for `conda` does not follow the same as `setup.py`.
Please change:
```  - numpy~=1.22.4```
with:
```  - numpy=1.22```

jlenain (changes requested):

mdpunch (commented):
Done

environment.yml
mdpunch: Done (I think)

jlenain:
> Done

Thanks a lot, @mdpunch , merging now.

PR #24, opened 2023-01-03
Update environment-dirac.yml

mdpunch:
With just - diracgrid::fts3 in the environment-dirac, this gives the error `ResolvePackageNotFound:  - diracgrid::fts3`. Patrick Sizun's solution to use `diracgrid/label/old::fts3` instead in the environment-dirac works, as might be guessed from https://anaconda.org/diracgrid/fts3

### DIFF
--- a/environment-dirac.yml
+++ b/environment-dirac.yml
@@ -10,7 +10,7 @@ dependencies:
   - pytest-runner
   - pip
   - dirac-grid  # required by DIRAC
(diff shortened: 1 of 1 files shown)


jlenain:
Hi @mdpunch !
Thanks a lot. As discussed on Slack, and already committed here, the `fts3` module dependency can now directly be installed via the `conda-forge` channel instead of pinning it to an old label from the `diracgrid` channel.
All good, will merge, many thanks !

PR #25, opened 2023-01-09
changes to improve exemple calibration script

frnbrun:
The modifications aim at improving the code, in particular by :
- improving masked array handling
- changing the parts related to event.trigger.event_type to match latest reader
- adding the possibility for the user to chose the actual trigger type of flat-field events to use

### DIFF
--- a/nectarchain/user_scripts/fbrun/ComputeCalibCoefs.py
+++ b/nectarchain/user_scripts/fbrun/ComputeCalibCoefs.py
@@ -55,13 +55,14 @@ class ncam_wfs:
         self.npixels = 0
         self.geometry =0
 
-    def load_wfs(self,nevents,file_name,ped_only=0,ff_only=0):
+    #ff_ttypes = trigger type of the FF events
+    def load_wfs(self,nevents,file_name,ped_only=0,ff_only=0,ff_ttypes=0):
         if nevents == 0:
             inputfile_reader = NectarCAMEventSource(input_url=file_name)
             for counter in enumerate(inputfile_reader):
                 nevents+=1
 #        else:
-        inputfile_reader = NectarCAMEventSource(input_url=file_name,max_events=nevents)
+        inputfile_reader = NectarCAMEventSource(input_url=file_name,max_events=nevents*10)
 
         print("N events request = ",nevents)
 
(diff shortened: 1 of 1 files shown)


jlenain:
Hi @frnbrun !

I will swiftly go through your PR and provide a quick, short review.

nectarchain/user_scripts/fbrun/ComputeCalibCoefs.py
```diff
@@ -88,24 +89,29 @@ def load_wfs(self,nevents,file_name,ped_only=0,ff_only=0):
         wfs_lg_tmp=np.zeros((1855,60))
```
jlenain, line 89: Instead of hardcoding the number of pixels and samples, maybe we should rely here on https://github.com/cta-observatory/ctapipe_io_nectarcam/blob/master/ctapipe_io_nectarcam/constants.py, e.g. with something like:

```python
from ctapipe_io_nectarcam import constants
[...]
wfs_lg_tmp = np.zeros((constants.N_PIXELS, constants.N_SAMPLE))

File: nectarchain/user_scripts/fbrun/ComputeCalibCoefs.py [hunk 2]
```diff
@@ -88,24 +89,29 @@ def load_wfs(self,nevents,file_name,ped_only=0,ff_only=0):
...
-            if ff_only and event.trigger.event_type != 4: continue
+            if ff_only and int(event.trigger.event_type.value) != ff_ttypes: continue
 
             self.wfs_evtid.append(event.index.event_id)
-            self.wfs_evttime[i]=event.nectarcam.tel[0].evt.ucts_timestamp
-            self.wfs_triggertype[i]=event.trigger.event_type
+            self.wfs_evttime[self.size]=event.nectarcam.tel[0].evt.ucts_timestamp
+            self.wfs_triggertype[self.size]=event.trigger.event_type.value
 
             trig_in= np.zeros((1855))
```
jlenain, line 109: Same here with e.g.:

```python
trig_in = np.zeros((constants.N_PIXELS))
```

nectarchain/user_scripts/fbrun/ComputeCalibCoefs.py
```diff
@@ -88,24 +89,29 @@ def load_wfs(self,nevents,file_name,ped_only=0,ff_only=0):
...
             for slice in range(4):
                 trig_in=np.logical_or(trig_in, event.nectarcam.tel[0].evt.trigger_pattern[slice])
-            self.trig_pattern[i]+=trig_in
+            self.trig_pattern[self.size]+=trig_in
+
+            #if ff_only and len(np.where(self.trig_pattern[self.size]>0)[0]) < 200: continue
 
             wfs_hg_tmp=np.zeros((1855,60))
             wfs_lg_tmp=np.zeros((1855,60))
```
jlenain, line 117: Same comment as above.

nectarchain/user_scripts/fbrun/ComputeCalibCoefs.py
```diff
@@ -191,12 +197,13 @@ def findrun_firstfile(run_number):
     run_number = sys.argv[2]
     ped_run_number = sys.argv[3]
     gain_run_number = sys.argv[4]
+    ff_ttypes = int(sys.argv[5])
```
jlenain, line 200: In order to avoid misuse of the code, could it be possible to add some checks of acceptable values for `ff_types` ?

jlenain (commented):
Hi @frnbrun,

I have provided some suggestions, possibly to be implemented or discussed.

jlenain (approved):

jlenain:
Hi @frnbrun !
Thanks a lot, can I merge now, or do you have other changes you would like to make ?

frnbrun:
That should be ok, although I noticed that my editor had line endings as CRLF for this file. It's better to have LF, no? 
If so, I can push another commit just changing this parameter.

PR #30, opened 2023-01-26
Bug fix in user script for calibration parameters estimation

frnbrun:
Fixes a bug - wrong boundary in a loop (leftover from tests)


jlenain (approved):
Allright, thanks @frnbrun !

frnbrun:
Thanks @jlenain !

PR #27, opened 2023-01-10
Fix conda env

jlenain:
The proposed PR, along with https://github.com/cta-observatory/ctapipe_io_nectarcam/pull/32, should address #26, and streamline the installation procedure of `nectarchain`.

PR #32, opened 2023-01-30
Changed DQM container build

jlenain:
In this PR, the singularity image version to be used for the DQM has been re-built against #27


jlenain:
Tested to work with run 3831, locally and on the EGI grid via DIRAC.

Issue #43, opened 2023-02-14
Incorrect time window for the charge extraction in DQM


jlenain:
While reviewing #42 , I saw a potential bug in the configuration of `LocalPeakWindowSum` used for the charge extraction in the Data Quality Monitoring module. The configuration is as follows:

```
        config = Config(
            {"LocalPeakWindowSum": {"window_shift": 12, "window_width": 12}}
```

The charge is then extracted over 12 ns, but with a time reference 12 ns before the actual pulse peak.

Could you please have a look, @hashkar ?

hashkar:
Yes @jlenain the values here are not right, they are the leftover of a very old example.  As I don't know what the exact values should be we can either leave it for the user to choose. If we now know what the values should be we can fix it once and for all. Otherwise we can discuss it during the software meeting.

jlenain:
Hi @hashkar !
Thanks for the reply.
A mechanism could indeed be introduced to let the user choose the time window to use.
However, in the meantime I would also propose to fix the current window to better encompass the pulse signal, e.g. [-4;+12] ns.

PR #45, opened 2023-02-20
Addresses #43

jlenain:
Change time window for charge extraction in DQM.


dkerszberg:
Thanks for the quick fix @jlenain. No problem with this PR I will merge it now.

PR #49, opened 2023-03-27
Black and flake8 formatting and adding flake8 in CI

sona-patel:
This PR formats the entire `nectarchain` according to black and flake8 style!


jlenain:
Hi @sona-patel ,
As discussed during the last NectarCAM scientific software meeting, sorry for overlooking this PR.
I would suggest to split the linting part from the integration into the CI. Concerning the latter, it would be best to simply use a GitHub action, such as:
https://github.com/py-actions/flake8
instead of calling `flake8` on each part of the code.

Concerning the former, could you please resolve the current merge conflicts ?

PR #50, opened 2023-03-29
Enable all ctapipe charge extractors

jlenain:
This PR enables all possible charge extractors currently implemented in [ctapipe](https://cta-observatory.github.io/ctapipe/ctapipe_api/image/extractor.html#image-extraction-methods).


guillaumegrolleron:
Indeed this modification should solve the issue, but I performed some tests and there is an error with numba for the charge extraction with the GlobalPeakWindowSum. I don't have solution yet but I will investigate before merging

PR #53, opened 2023-04-07
Apptainer CI

jlenain:
This PR introduces a new CI workflow where:

- `nectarchain` is shipped as a `conda` environment, and buit as an Apptainer/Singularity container on selected events (releases, pull requests, merges into `master`),
- the container is deployed on the GitHub Container Registry, i.e. visible as a package of the cta-observatory organization, on non-pull-request events, such as releases.

Users can then use such containers with:
```shell
apptainer shell oras://ghcr.io/cta-observatory/nectarchain:latest
```

**TL;DR**

We will now have fully automatic Apptainer containers built and deployed on releases!

@kosack , @maxnoe , would that be deemed acceptable ?

### DIFF
--- a/singularity/Singularity
+++ b/singularity/Singularity
@@ -14,6 +14,12 @@
 Bootstrap: docker
 From: condaforge/mambaforge
 
+%setup
+    mkdir -p ${SINGULARITY_ROOTFS}/opt/cta/nectarchain
+
+%files
+    . /opt/cta/nectarchain
+
 # From https://github.com/hpcng/singularity/issues/5075#issuecomment-594391772
 %environment
     action="${0##*/}"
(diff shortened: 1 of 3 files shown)


singularity/Singularity
```diff
@@ -47,7 +44,8 @@ From: condaforge/mambaforge
     cd /opt/cta
 
     # Install nectarchain
-    mamba env create --quiet --file /tmp/environment.yml --prefix /opt/conda/envs/nectarchain
+    git clone https://github.com/cta-observatory/nectarchain.git
```
maxnoe: Cloning inside the singularity file means that you always build the default branch. So your CI tests wouldn't actually catch the case where a PR breaks the singularity build. 

I think you should add the local files.

jlenain: You're absolutely right, thanks a lot! That was actually what I also intended to do at the beginning, but I forgot along the way that everything needed is already there thanks to the checkout action :sweat_smile:.

jlenain: @maxnoe , should be properly done now

Issue #54, opened 2023-04-11
HDU names in fits containing extracted waveform and charge


sona-patel:
Output fits files generated from `load_wfs_compute_charge.py` do not have names for any of the hdus (eg. shown below for charge_run3936.fits). 

<img width="821" alt="charge_run3936" src="https://user-images.githubusercontent.com/16776228/231182392-3ebfc4e8-4e04-43fe-97da-8925f1987d1c.png">

Wouldn't it be nice to give them a meaningful name? 

Thanks!

jlenain:
Thanks for reporting this, @sona-patel .
@guillaumegrolleron , could you please have a look ?

guillaumegrolleron:
Hi @sona-patel , @jlenain , 
#59 is answering this issue

jlenain:
Hi !
Thanks a lot, @guillaumegrolleron !
@sona-patel , here are the resulting extension names:
![Screenshot from 2023-04-27 22-57-05](https://user-images.githubusercontent.com/5399958/234989479-3fd5a3fd-89da-4938-9e1b-7ef6c3f4d964.png)
![Screenshot from 2023-04-27 22-57-56](https://user-images.githubusercontent.com/5399958/234989620-a4e8746a-7403-437c-bed7-37dee9a96ff8.png)

jlenain:
Solved in #59

PR #59, opened 2023-04-25
Names for charges and waveforms container extension in fits files

guillaumegrolleron:
This PR is to answer the request #54. Now extension names in the fits file of saved ChargesContainer and Waveforms Container are describing the data within. 

There are also some minor bugfix


jlenain (approved):
Thanks a lot, @guillaumegrolleron !

PR #69, opened 2023-06-16
py-actions/flake8 added in ci

sona-patel:


jlenain (approved):
Thanks a lot, @sona-patel !

PR #72, opened 2023-06-28
DQM database and web application

jlenain:
This PR introduces the web application for the Data Quality Monitoring tool, as well as all the needed recipes to setup the accompanying infrastructure (database configuration, recipes to create the two needed virtual machine on a local cloud with e.g. OpenStack).


tibaldo:
Would it be too much to ask to add a test routine for this new material before the PR is merged?

jlenain:
> Would it be too much to ask to add a test routine for this new material before the PR is merged?

Hi @tibaldo,

It would be tough to implement extensive unit tests without actually deploying the overall VM infrastructure, but I can certainly implement some, at least for the data base utilities and maybe on the bokeh side.

jlenain:
Hi @tibaldo ,

I think I found a way to extensively test all these new features, using an in-memory data base for unit tests. This PR is ready for another review.

src/nectarchain/dqm/bokeh_app/main.py
```diff
@@ -0,0 +1,77 @@
...
+
+from nectarchain.dqm.db_utils import DQMDB
+from app_hooks import get_rundata, make_camera_displays
+
+NOTINDISPLAY = ['Results_TriggerStatistics',
+                'Results_MeanWaveForms_HighGain',
+                'Results_MeanWaveForms_LowGain',
+                'Results_CameraMonitoring']
+
+geom = CameraGeometry.from_name("NectarCam-003")
```
tibaldo, line 22: Should the camera geometry to use be a configurable parameter?

src/nectarchain/dqm/bokeh_app/tests/test_app_hooks.py
```diff
@@ -0,0 +1,62 @@
...
+# bokeh imports
+from bokeh.layouts import layout, row
+from bokeh.models import Select
+from bokeh.plotting import curdoc
+from bokeh.io import output_file, save
+
+from ctapipe.instrument import CameraGeometry
+from ctapipe.coordinates import EngineeringCameraFrame
+
+geom = CameraGeometry.from_name("NectarCam-003")
```
tibaldo, line 13: Same comment as above

tibaldo (approved):
Hi @jlenain, thanks a lot for this impressive amount of work! Using the ZEO/ZODB servers and running tests via an in-memory instance looks like a very clever and effective solution.
I only have two minor comments on hard-coding the camera geometry to be used, but otherwise the PR looks good to be merged to me.

src/nectarchain/dqm/bokeh_app/main.py
jlenain: Or I guess, we could just use a more generic `NectarCam` parameter, which should work.

src/nectarchain/dqm/bokeh_app/app_hooks.py
```diff
@@ -13,7 +13,7 @@
                 'Results_MeanWaveForms_LowGain',
                 'Results_CameraMonitoring']
 
-geom = CameraGeometry.from_name("NectarCam-003")
+geom = CameraGeometry.from_name("NectarCam")
```
tibaldo, line 16: I'm not sure this actually works as intended. I suspect it does not load a generic Geometry, but an older version preceding 003. I was rather suggesting to add a single parameter somewhere that can set the geometry all across the package.

vmarandon: Perhaps it would be better, in this case, to have this line in a function that could be used everywhere.
Something like : 
```
def GetCameraGeometry(name="NectarCam-003"):  
      return CameraGeometry.from_name(name)
```

jlenain: Agree with @vmarandon , I was thinking about the same: maybe even easier to have a constant for the camera geometry version across the whole package, in something like `utils/constants.py`.

I will revert the last change and propose to address this in another PR, we planned to have one on re-organization of the code anyway.

kosack: If you have a working EventSource, you can also take it's `SubarrayDescription` (`source.subarray`) and write it to an HDF5 file, and load that when needed. E.g.:
```py3
with EventSource("some_nectarcam_file.zfits.gz")  as source:
    source.subarray.to_hdf("subarray.h5")

# then later:

subarray = SubarrayDescription.from_hdf("subarray.h5")
geom = subarray.tel[1].camera.geometry
```

jlenain: Thanks a lot, @kosack .
Indeed, when working with an active EventSource, I agree using `source.subarray.tel[0].camera.geometry` is the most robust and self-consistent way to get the correct camera geometry.

In this PR, though, the context for the running Bokeh application is independent from any EventSource, the input data are fed from an object-oriented data base. They are stored as nested dictionaries there, and so we "lost" at that point part of the original meta-data. Just for visualization purpose, this was a way to correct the geometry right in camera displays.

kosack: I guess it's only by luck though that the simulated pixel ids are the same as the real ones? That is not the case in general. So just beware if we switch simulations, it could "break" images in such a display.

PR #76, opened 2023-09-12
Updated scripts

vmarandon:


src/nectarchain/user_scripts/vmarandon/Utils.py
```diff
@@ -368,3 +426,17 @@ def GetEventTypeFromString(event_str):
         print(f"WARNING> Don't know about the event type [{event_str}]")
         evt_type = EventType.UNKNOWN
     return evt_type
+
+
+def save_simple_data(datas, fileName):
+    """Save info in a file using pickle"""
+    with lz4.frame.open( fileName, 'wb' ) as f :
+        pickle.dump(datas,f)
```
jlenain, line 434: I tend to recommend not using pickles, for security reasons, as they can hide malicious code snippets. Is there a good reason to use them here ?

vmarandon: This pickle usage is not meant to be used routinely for nectarchain. I would avise also against it in the "regular" code and recommend using derived class from Container of ctapipe.
Here, it's just to keep the code in line with the modification I've done to my scripts to perform some SPE analysis as I wanted to be fast (ICRC emergency).  In this case, pickle is quite convenient to avoid having to write I/O code, especially when you want to pass data from one notebook to another to play with.

jlenain: All right, we're on the same page!

jlenain (approved):

PR #74, opened 2023-07-28
Lint code

jlenain:
This PR proposes to lint the entire repository (except user scripts), using the rules for `black` and `flake8` defined in our pre-commit hooks (in `.pre-commit-config.yaml`), and introduced in #71.

At this moment, only `nectarchain.calibration.container.charge_extractor` is missing, as I would like to clarify a point in the `__call__` method with @guillaumegrolleron beforehand.

PR #78, opened 2023-09-15
Refactoring of the calibration and container module

guillaumegrolleron:
I have been working last couple of days on the refactoring of the calibration code. As discussed with @jlenain and @vmarandon, we would like to have one module (`makers`) to make computations on raw data and give the calibrations coefficient (or whatever). Then the other modules should use the `makers` and all the computation part would be in the same place.  
In this PR, I tried to achieve that goal. The container module has been completely rebuilt. Now the `ChargesContainers` and `WaveformsContainers` are only made to store data. This classes inherit from the `ctapipe.container `. The computation part which was previously in the `container` module has been moved in the `makers` module into two classes, the `WaveformsMaker`  which is basically looping through an `EventSource` and stacking events informations and waveforms into a `WaveformsContainers`. The `ChargesMaker` do the same but extracting the charge instead of the waveforms. 
The calibration part is now into the `makers` module. The SPE fit classes have been rebuilt to speed up the process. 
I also started to write unit tests, but the coverage is not complete.


guillaumegrolleron:
I all, 
I removed in pytest the tests which needs run files to be performed. 
But I have been looking into `lstchain`, and they use a [(script)](https://github.com/cta-observatory/cta-lstchain/blob/main/download_test_data.sh) to download from an external server with a password provided to the repository through the GitHub secrets [feature](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions). I guess we should easily do the same.

jlenain:
Thanks a lot, @guillaumegrolleron , for all the work.
I would propose to merge everything, and we will learn altogether how to use the `ctapipe` components.
Let me first resolve the current conflicts with `master` (meaning, first deleting ` src/nectarchain/calibration/container/charge.py` and ` src/nectarchain/calibration/container/waveforms.py` before merging).

jlenain (approved):

jlenain:
While linting the code introduced in this PR via #74 , it seems some parts of @guillaumegrolleron 's refactoring of the code were not yet included in this PR.

I am very sorry for this, it seems we merged this PR a bit too fast.
Nevermind, let's have the rest in a new PR.

Pinging @guillaumegrolleron and @vmarandon .

PR #79, opened 2023-10-17
Modification to `WaveformContainer` and `ChargeContainer` for SPE-WT

sona-patel:
This PR modifies the `WaveformContainer` and `ChargeContainer` (mainly `load_wfs` and `write` function) to read and store waveform given the list of event_id for splitting.

### DIFF
--- a/src/nectarchain/calibration/container/charge.py
+++ b/src/nectarchain/calibration/container/charge.py
@@ -183,11 +183,14 @@ class ChargeContainer:
 
         return chargeContainer
 
-    def write(self, path: Path, bloc, start, end, **kwargs):
+    def write(self, path: Path, start=None, end=None, block=None, **kwargs):
         """method to write in an output FITS file the ChargeContainer.
 
         Args:
             path (str): the directory where you want to save data
+            start (int): Start of event_id. Default is None.
+            end (int): End of event_id. Default is None.
+            block (int): Block number of SPE-WT scan. Default is None.
         """
         suffix = kwargs.get("suffix", "")
         if suffix != "":
@@ -282,7 +285,7 @@ class ChargeContainer:
         )
         try:
             hdul.writeto(
-                Path(path) / f"charge_run{self.run_number}{suffix}_bloc_{bloc}.fits",
+                Path(path) / f"charge_run{self.run_number}{suffix}_bloc_{block}.fits",
                 overwrite=kwargs.get("overwrite", False),
             )
             log.info(
--- a/src/nectarchain/calibration/container/charge.py
+++ b/src/nectarchain/calibration/container/charge.py
@@ -1,70 +1,52 @@
+import glob
+import logging
+import os
+import time
 from argparse import ArgumentError
+from pathlib import Path
+
 import numpy as np
 import numpy.ma as ma
-from matplotlib import pyplot as plt
-import copy
-from pathlib import Path
-import glob
-import time
-import sys
-import os
-import logging
-logging.basicConfig(format='%(asctime)s %(name)s %(levelname)s %(message)s')
-log = logging.getLogger(__name__)
-log.handlers = logging.getLogger('__main__').handlers
-
-from ctapipe.visualization import CameraDisplay
-from ctapipe.coordinates import CameraFrame,EngineeringCameraFrame
-from ctapipe.instrument import CameraGeometry
-from ctapipe.image.extractor import (FullWaveformSum,
-    FixedWindowSum,
-    GlobalPeakWindowSum,
-    LocalPeakWindowSum,
-    SlidingWindowMaxSum,
-    NeighborPeakWindowSum,
-    BaselineSubtractedNeighborPeakWindowSum,
-    TwoPassWindowSum)
-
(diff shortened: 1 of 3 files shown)


jlenain:
Thanks, @sona-patel !
Is this PR still in the draft stage ? The title seems to indicate so.

src/nectarchain/calibration/container/charge.py
```diff
@@ -124,140 +126,212 @@ def __init__(self,charge_hg,charge_lg,peak_hg,peak_lg,run_number,pixels_id,neven
...
         chargeContainer.event_type = waveformContainer.event_type
         chargeContainer.event_id = waveformContainer.event_id
         chargeContainer.trig_pattern_all = waveformContainer.trig_pattern_all
 
-        return chargeContainer 
+        return chargeContainer
 
-
-    def write(self,path : Path,**kwargs) : 
+    def write(self, path: Path, bloc, start, end, **kwargs):
```
jlenain, line 186: I think `bloc` (`block` ?), `start` and `end` should have default values such that a whole run is processed and written if no value are provided for these arguments.

src/nectarchain/calibration/container/charge.py
```diff
@@ -124,140 +126,212 @@ def __init__(self,charge_hg,charge_lg,peak_hg,peak_lg,run_number,pixels_id,neven
...
+        hdr["NPIXELS"] = self.npixels
+        hdr["COMMENT"] = (
+            f"The charge containeur for run {self._run_number} with {self._method} "
+            f"method : primary is the pixels id, then you can find HG charge, "
+            f"LG charge, HG peak and LG peak, 2 last HDU are composed of event "
+            f"properties and trigger patern"
+        )
+
+        primary_hdu = fits.PrimaryHDU(self.pixels_id, header=hdr)
+        charge_hg_hdu = fits.ImageHDU(self.charge_hg[start:end], name="HG charge")
```
jlenain, line 227: For instance here, the default behavior should allow for the full `charge_hg` attribute to be stored and written to disk.

src/nectarchain/calibration/container/waveforms.py
```diff
@@ -136,138 +157,196 @@ def check_events(self):
...
+                    f"trigger type"
+                )
+            n_events += 1
         return n_events
 
-
-
-    def load_wfs(self,compute_trigger_patern = False):
-        """mathod to extract waveforms data from the EventSource 
+    def load_wfs(self, start, end, compute_trigger_patern=False):
```
jlenain, line 172: A similar comment here, I think `start` and `end` should have default values such that the all events should be loaded.

src/nectarchain/calibration/container/waveforms.py
```diff
@@ -136,138 +157,196 @@ def check_events(self):
...
-            self.ucts_timestamp[i]=event.nectarcam.tel[WaveformsContainer.TEL_ID].evt.ucts_timestamp
-            self.event_type[i]=event.trigger.event_type.value
-            self.ucts_busy_counter[i] = event.nectarcam.tel[WaveformsContainer.TEL_ID].evt.ucts_busy_counter
-            self.ucts_event_counter[i] = event.nectarcam.tel[WaveformsContainer.TEL_ID].evt.ucts_event_counter
+        wfs_hg_tmp = np.zeros((self.npixels, self.nsamples), dtype=np.uint16)
+        wfs_lg_tmp = np.zeros((self.npixels, self.nsamples), dtype=np.uint16)
 
+        # n_traited_events = 0
+        event_start, event_end = start, end
+        print(event_start, event_end)
```
jlenain, line 187: Plain prints should be avoided, to the benefit of the logging system.

src/nectarchain/calibration/container/waveforms.py
```diff
@@ -136,138 +157,196 @@ def check_events(self):
...
 
-        #if compute_trigger_patern and np.max(self.trig_pattern) == 0:
+        # if compute_trigger_patern and np.max(self.trig_pattern) == 0:
         #    self.compute_trigger_patern()
 
-
-    def write(self,path : str, **kwargs) : 
-        """method to write in an output FITS file the WaveformsContainer. Two files are created, one FITS representing the data
-        and one HDF5 file representing the subarray configuration
+    def write(self, start, end, bloc, path: str, **kwargs):
```
jlenain, line 227: Here also, `start`, `end`, and `bloc` (`block` ?) should default to the whole run.

jlenain (changes requested):
Thanks for this, @sona-patel . I think the class behaviors should default to loading/writting whole runs. In such a case, default values should be provided in the loading/writting methods.

sona-patel:
> Thanks, @sona-patel ! Is this PR still in the draft stage ? The title seems to indicate so.

Thanks for reviewing @jlenain!
Yes, I kept it as a draft, since I also want to add the user_scripts using this case. What do you suggest? Should I put it along with this PR or in a separate PR?

src/nectarchain/calibration/container/waveforms.py
sona-patel: Agreed! Its remnant from the notebook 😀!

jlenain (approved):

PR #63, opened 2023-06-13
DQM: Adding protections for runs with 21 masked pixels (due to shutter)

hashkar:
fixing the 21 pixels issues for old runs: addin a check first.

PR #80, opened 2023-10-20
DQM: Adding protections for runs with 21 masked pixels (due to shutter)

hashkar:
The camera geometry was already fixed. This pull request has small bug and style fixes


jlenain:
Was a follow-up of #63

PR #88, opened 2024-01-23
Simple script to retrieve a run from the grid (and the related db info)

vmarandon:


jlenain (approved):
Thanks a lot, @vmarandon !
For the `DataManagement` class, pour idea with @guillaumegrolleron was to use the DIRAC API directly, and avoid system calls using subprocesses. However, your script comes very handy!

PR #90, opened 2024-01-25
Find runs on GRID directly with the Dirac python API

guillaumegrolleron:
I just changed the way to access the runs files stored on the GRID, before we had to access the ELog and the method was not robust. 

Now we directly use the DIRAC API to search and find location of data with run number.

This feature makes possible to use `tools` and do not need to previously download runs files locally in $NECTARCAMDATA. `nectarchain` will check if the files are accessible locally, if not, it will automatically fetch them from the GRID and store them in ¤NECTARCAMDATA

### DIFF
--- a/src/nectarchain/data/management.py
+++ b/src/nectarchain/data/management.py
@@ -8,15 +8,12 @@ log.handlers = logging.getLogger("__main__").handlers
 import glob
 import os
 import pathlib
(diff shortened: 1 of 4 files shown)


jlenain (approved):
That's brilliant, @guillaumegrolleron ! Thanks a lot !

PR #115, opened 2024-03-21
Add container/component/tool for pedestal estimation

tibaldo:
This PR add a tool/component dedicated to estimate pedestals based on data that contain events of type `SKYPEDESTAL`. Pedestals are stored in a dedicated container following the nectarchain conventions (`pixels_id` etc.).
Several methods for filtering the data are available based on timestamps of the events, standard deviations of the waveforms and the charge distribution.
In order to reduce the memory load the tool is built in such a way that pedestals are estimated in slices with a number of events `events_per_slice` and then results combined at the end.
PS: I'm not sure why in the file comparison I see the renaming master-> main which is already integrated in the upstream repository


jlenain:
CI tests are currently failling due to #116 . #117 should fix it.

jlenain:
> PS: I'm not sure why in the file comparison I see the renaming master-> main which is already integrated in the upstream repository
I think it is, maybe, because you pull from the main branch of the repo into your dev branch before pushing your commits into it ?
If so, new features introduced in the main are duplicated in your dev branch. This should not be any issue at merge (we'll clean the commit log), but the proper workflow is to ignore any new commit from the main in your dev branch. Any potential conflicts with the main will be flagged, and can be fixed, when we merge the PR.

jlenain (approved):
Thanks a lot, @tibaldo !
I tested your branch locally, and everything is running very smoothly.
To me, the PR is in an excellent shape and is ready to be merged.

PR #122, opened 2024-07-03
Thermal tests + CI in DQM

hashkar:
This pull request contains changes to make the DQM easily usable for the thermal tests.
The main changes also include the integration of CI tests for the DQM.
This is just an initiation to the CI tests. They are not complete yet. The first function of each class is tested here.


hashkar:
One thing that needs to be fixed before accepting this pull request is the path to the data file used for the tests. @jlenain can you point me to the one you mentioned yesterday?

jlenain:
> One thing that needs to be fixed before accepting this pull request is the path to the data file used for the tests. @jlenain can you point me to the one you mentioned yesterday?

There are a few such instances throughout `nectarchain`, e.g. https://github.com/cta-observatory/nectarchain/blob/e72010a698523a5fa410546df9efe567d7229a4b/src/nectarchain/makers/calibration/tests/test_pedestal_tool.py#L11

jlenain:
@hashkar , one thing that I am wondering is whether the DQM module is the best place for thermal test analyses. In my understanding, the DQM should run a quality monitoring as generic as possible. A dedicated analysis of runs for thermal tests would be better placed into `user_scripts`, in my opinion. What do you think ? @frnbrun , @vmarandon as well ?

hashkar:
@jlenain I totally agree. The DQM module is still the same I just changed small things in it to be able to compute the quantities needed for the thermal tests easily. These quantities are also useful for the DQM (for example pedestal averaged for the first 20 samples). In the user_script there will be the script to launch the thermal tests and analyze + plot the results.

jlenain:
I guess the prints were meant to assess which value to assert the tests. Could you please remove these? Thanks!

environment.yml
```diff
@@ -15,6 +15,8 @@ dependencies:
   - pandas
   - zeo
   - zodb
+  - openssl
+  - libexpat
```
jlenain, line 19: I am reluctant to add explicit dependencies to such standard packages, which implicitly comes with other dependencies. This may impact our conda package builds.

src/nectarchain/dqm/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .db_utils import DQMDB
 from .dqm_summary_processor import DQMSummary
+from .pixel_participation import PixelParticipationHighLowGain
 
-__all__ = ["DQMDB", "DQMSummary"]
+__all__ = ["DQMDB", "DQMSummary", "PixelParticipationHighLowGain"]
```
jlenain, line 5: What is the reason why `PixelParticipationHighLowGain` should be exposed, and not the other DQM classes ? We should take care of the fact that PR #148 already exposes all other DQM classes. We may hit a conflict at some point, which should be easy to solve.

src/nectarchain/dqm/camera_monitoring.py
```diff
@@ -4,11 +4,11 @@
 import numpy as np
 from astropy import time as astropytime
 from ctapipe.coordinates import EngineeringCameraFrame
-from ctapipe.instrument import CameraGeometry
 from ctapipe.visualization import CameraDisplay
-from dqm_summary_processor import DQMSummary
 from matplotlib import pyplot as plt
 
+from nectarchain.dqm.dqm_summary_processor import DQMSummary
+
```
jlenain, line 11: Covered in https://github.com/cta-observatory/nectarchain/pull/148/files#diff-593f6bdf396b1884f5f268dc53b094157f91cbb60ad009414793c0a40c143cde

src/nectarchain/dqm/charge_integration.py
```diff
@@ -2,12 +2,12 @@
...
 from ctapipe.coordinates import EngineeringCameraFrame
 from ctapipe.image import LocalPeakWindowSum
-from ctapipe.instrument import CameraGeometry
 from ctapipe.visualization import CameraDisplay
-from dqm_summary_processor import DQMSummary
 from matplotlib import pyplot as plt
 from traitlets.config.loader import Config
 
+from nectarchain.dqm.dqm_summary_processor import DQMSummary
+
```
jlenain, line 10: Covered in PR #148

src/nectarchain/dqm/mean_waveforms.py
```diff
@@ -1,7 +1,8 @@
 import numpy as np
-from dqm_summary_processor import DQMSummary
 from matplotlib import pyplot as plt
 
+from nectarchain.dqm.dqm_summary_processor import DQMSummary
+
```
jlenain, line 5: Covered in PR #148

jlenain (approved):
Some changes may be redundant with PR #148 , or adjusted accordingly. We'll take care of that when merging the two PRs.

PR #123, opened 2024-07-03
Split on event type

vmarandon:
Add a DelimiterLoopNectarCAMCalibrationTool class that inherits nearly everything from EventsLoopNectarCAMCalibrationTool that give the possibility to split slices with the UNKNOWN event type. This is used in some runs to actually indicate some changes in acquisition conditions (like X-Y table position change, etc...)

### DIFF
--- a/src/nectarchain/makers/core.py
+++ b/src/nectarchain/makers/core.py
@@ -34,7 +34,7 @@ from ..utils import ComponentUtils
 from .component import *
 from .component import NectarCAMComponent, get_valid_component
 
-__all__ = ["EventsLoopNectarCAMCalibrationTool"]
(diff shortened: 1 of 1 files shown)


vmarandon:
@jlenain , @guillaumegrolleron , @frnbrun : This is a draft to add a Tool that could split slices when there is a certain trigger type. It's still a draft as it needs to be tested :-D

jlenain (approved):

jlenain:
Thanks a lot, @vmarandon !
One thing I am wondering: when a run is sliced when encountering an UNKOWN event type, is this event skipped or included in the corresponding slice ? If so, at the start or the end of the slice ?

PR #129, opened 2024-07-25
Resolve conflict between python==3.11 and numba~=0.56.0 for installation on osx-arm64

### LINKED ISSUE #126: Conflict between python==3.11 and numba~=0.56.0 for installation on osx-arm64
I'm trying to install nectarchain on a new laptop with a Mac ARM chip and I get an incompatibility issue. On this platform numba~=0.56.0 is only compatible with python<3.11, and python 3.1 requires numba >=0.60. Is is possible to relax these version constraints or otherwise is there a known solution to this bug?

jlenain:
This PR should fix #126.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,10 @@ jobs:
             python-version: "3.11"
             ctapipe-version: "v0.19.3"
             install-method: "mamba"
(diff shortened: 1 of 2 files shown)


tibaldo (approved):
Everything looks good. Thank you @jlenain

Issue #137, opened 2024-07-29
Lint does not run in CI


jlenain:
**Describe the bug**
The code is not linted in CI.

**To Reproduce**
1. Open a PR
2. Check the runner logs
3. All linting steps are skipped

**Expected behavior**
The pre-commit hooks should be run in CI

**Supporting information**
See e.g. https://github.com/cta-observatory/nectarchain/actions/runs/10144120767/job/28046944606
```
[...]
Run pre-commit run --show-diff-on-failure --color=always --files $(git diff origin/main --name-only)
fatal: ambiguous argument 'origin/main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
isort................................................(no files to check)Skipped
black-jupyter........................................(no files to check)Skipped
flake8...............................................(no files to check)Skipped
jupytext.............................................(no files to check)Skipped
```

PR #140, opened 2024-07-31
Fixing lint in CI

jlenain:
This PR should solve #137


jlenain:
Tested that it works as intended by forcing a faulty coding style in https://github.com/jlenain/nectarchain/actions/runs/10177628398/job/28149523371

vmarandon (approved):

PR #141, opened 2024-07-31
Automatic documentation

jlenain:
This PR extends #130 by including a full automatic documentation of the whole API, just by using the `recursive` option of `sphinx.ext.autosummary`.

It also fixes quite a few typos in Docstrings, to comply with the reStructuredText format.


jlenain:
The docs job in CI fails, because of a misformatted docstring in `ctapipe_io_nectarcam`. As soon as we also document the plugin, this failure should go away, by mapping the `ctapipe_io_nectarcam` documentation through an `intersphinx-mapping` instruction.

vmarandon (approved):

PR #147, opened 2024-08-02
DQM output format and parsing script

jlenain:
This PR introduces a new format for the DQM output in FITS files, as well as a user script which, once a DQM analysis has been run on DIRAC, grabs the results from DIRAC, extracts the archive and ingests the results in the ZODB data base.

This is a fundamental brick towards an automatic pipeline which would launch the DQM systematically on DIRAC on data transfer from CEA to DIRAC, and results ingestion in ZODB.

This was tested to run fine locally (see http://lpnhe-cta2.in2p3.fr/nectarcam-dqm), but we need to merge this PR, to trigger the building of an Apptainer image in CI, which would then be used to launch a couple of DQM jobs on DIRAC, to be able to further test the user script from start to finish.

In case of issues along the way, another PR will be opened.

### DIFF
--- a/src/nectarchain/dqm/dqm_summary_processor.py
+++ b/src/nectarchain/dqm/dqm_summary_processor.py
@@ -1,3 +1,4 @@
+import numpy as np
 from astropy.io import fits
 from astropy.table import Table
 
@@ -38,7 +39,13 @@ class DQMSummary:
         try:
             data[name] = content
         except TypeError:
-            data = Table(content)
+            try:
+                data = Table(content)
+            except ValueError:
+                # We may have caught just a single float value, try to pack it into
+                # the FITS output
+                content = np.array([content])
+                data = Table(content)
         hdu = fits.BinTableHDU(data)
         hdu.name = name
         return hdu
--- a/src/nectarchain/dqm/dqm_summary_processor.py
+++ b/src/nectarchain/dqm/dqm_summary_processor.py
@@ -40,7 +40,7 @@ class DQMSummary:
         hdu, hdu0, hdu1, hdu2 = None, None, None, None
         hdulist = fits.HDUList()
         for i, j in DICT.items():
-            if (i == "Results_TriggerStatistics"):
+            if i == "Results_TriggerStatistics":
                 for n2, m2 in j.items():
                     data2[n2] = m2
                     hdu2 = fits.BinTableHDU(data2)
--- a/src/nectarchain/dqm/dqm_summary_processor.py
+++ b/src/nectarchain/dqm/dqm_summary_processor.py
@@ -32,47 +32,22 @@ class DQMSummary:
     ):
         print("Processor 5")
 
-    def WriteAllResults(self, path, DICT):
-        data2 = Table()
-        data1 = Table()
-        data0 = Table()
+    @staticmethod
+    def _create_hdu(name, content):
         data = Table()
-        hdu, hdu0, hdu1, hdu2 = None, None, None, None
+        data[name] = content
+        hdu = fits.BinTableHDU(data)
+        hdu.name = name
+        return hdu
+
+    def WriteAllResults(self, path, DICT):
         hdulist = fits.HDUList()
         for i, j in DICT.items():
-            if i == "Results_TriggerStatistics":
-                for n2, m2 in j.items():
-                    data2[n2] = m2
-                    hdu2 = fits.BinTableHDU(data2)
-                    hdu2.name = n2
-                    hdulist.append(hdu2)
-
-            elif (i == "Results_MeanWaveForms_HighGain") or (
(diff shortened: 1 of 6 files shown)


src/nectarchain/dqm/dqm_summary_processor.py
```diff
@@ -40,50 +40,39 @@ def WriteAllResults(self, path, DICT):
...
+            elif (i == "Results_PixelTimeline_HighGain") or (
+                i == "Results_PixelTimeline_LowGain"
+            ):
                 for n0, m0 in j.items():
-                    data0[n0] = m0 
-                hdu0 = fits.BinTableHDU(data0)
-                hdu0.name = "BPX"
+                    data0[n0] = m0
+                    hdu0 = fits.BinTableHDU(data0)
+                    hdulist.append(hdu0)
```
vmarandon, line 65: The name is not set, is that normal ? 
I would expect to see the line : `hdu0.name = n0` or something like that.

jlenain: Ah, you're absolutely right.
And I think I found a way to simplify this method, see last commit.

vmarandon: Looks much better indeed. 

The change imply that now you have one `BinTableHDU` per result instead of 2 informations in one as I understand the code for some of the if cases.  Some names will also changed as some were hard coded in the previous version. Will it be a problems for code coming after the creation of the fits ?

jlenain: Actually, yes, it will. The DQM itself fails with such output structure. I should revert this.

vmarandon: It's weird because the "else" part of the old code is doing what you are doing in the "new" static function you are using. Maybe it should not be reverted but things can be fixed ? (I find more appealing to not have the long list of if/else :-) )

jlenain: Hi @vmarandon ,
I am currently working on that. The current error reads:
```python
Traceback (most recent call last):
  File "/opt/cta/nectarchain/src/nectarchain/dqm/start_dqm.py", line 243, in <module>
    p.WriteAllResults(ResPath, NESTED_DICT)
  File "/opt/cta/nectarchain/src/nectarchain/dqm/dqm_summary_processor.py", line 49, in WriteAllResults
    hdu = self._create_hdu(name, content)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/cta/nectarchain/src/nectarchain/dqm/dqm_summary_processor.py", line 40, in _create_hdu
    data[name] = content
    ~~~~^^^^^^
  File "/opt/conda/envs/nectar-dev/lib/python3.11/site-packages/astropy/table/table.py", line 2092, in __setitem__
    self.add_column(value, name=item, copy=True)
  File "/opt/conda/envs/nectar-dev/lib/python3.11/site-packages/astropy/table/table.py", line 2395, in add_column
    raise TypeError("Empty table cannot have column set to scalar value")
TypeError: Empty table cannot have column set to scalar value
```
as we are hitting something like "mixin" columns in `astropy.table`: https://docs.astropy.org/en/stable/table/mixin_columns.html#registering-array-like-objects-as-mixin-columns. I am looking for a solution avoiding the need to implement such a wrapper class.

jlenain: This boils down to the fact that the `TRIGGER-STATISTICS` and `START-TIMES` fields are just strings, and as such, cannot be stored as columns in `astropy.table`. The commit 95491ab324e9ea9432324471ada88c58715a56ac worked, but at the expanse of duplicated columns in the FITS output, as we noted with @hashkar.
Avoiding this duplication in a4dd3c47dcc8571dfd4a2d3c305855c2dd9974c0 triggered the issue discussed here.
I would see two solutions:
- simply ignoring the `TRIGGER-STATISTICS` and `START-TIMES` fields, which is not elegant, with something like:
```python
    def WriteAllResults(self, path, DICT):
        hdulist = fits.HDUList()
        for i, j in DICT.items():
            for name, content in j.items():
                try:
                    hdu = self._create_hdu(name, content)
                    hdulist.append(hdu)
                except TypeError as e:
                    print(f"DEBUG JPL: Got error {e}, skipping {name}")
                    pass
```
- or changing their format and storing them as dicts in https://github.com/cta-observatory/nectarchain/blob/main/src/nectarchain/dqm/trigger_statistics.py.

I'd go for the 2nd one.

jlenain: ... which gives the same issue...

jlenain: OK, I think I found a way.

PR #161, opened 2024-11-25
Flat field

ArmelleJB:
Add component, container and tool for calculation of flat field coefficients.

### DIFF
--- a/src/nectarchain/data/container/flatfieldContainer.py
+++ b/src/nectarchain/data/container/flatfieldContainer.py
@@ -3,3 +3,83 @@ import logging
 logging.basicConfig(format="%(asctime)s %(name)s %(levelname)s %(message)s")
 log = logging.getLogger(__name__)
 log.handlers = logging.getLogger("__main__").handlers
+
+import numpy as np
+from ctapipe.containers import Field
+from ctapipe.core.traits import ComponentNameList, Dict, Integer, List, Tuple
+
+from .core import NectarCAMContainer
+
+__all__ = ["FlatFieldContainer"]
+
+
+class FlatFieldContainer(NectarCAMContainer):
+    """
+    Container that holds flat field coefficients and other useful information
+
+    Fields:
+        run_number (np.uint16): Number of the run
+        npixels (np.uint16): Number of pixels
+        pixels_id (np.ndarray): Array of pixel's ID
+        ucts_timestamp (np.ndarray) : Array of time stamps of each event (UTC)
+        event_type (np.ndarray): Array of trigger event types (should be all flat field events)
+        event_id (np.ndarray): Array of the IDs of each event
+        amp_int_per_pix_per_event (np.ndarray): Array of integrated amplitude of each pulse
+        t_peak_per_pix_per_event (np.ndarray): Array of samples containing the pulse maximum
+        FF_coef (np.ndarray): Array of flat field coefficients
+        bad_pixels (List): List of pixel identified as outliers
+    """
+
+    run_number = Field(
+        type=np.uint16,
+        description="run number associated to the waveforms",
+    )
+
+    npixels = Field(
+        type=np.uint16,
+        description="number of effective pixels",
+    )
+
+    pixels_id = Field(type=np.ndarray, dtype=np.uint16, ndim=1, description="pixel ids")
+
+    ucts_timestamp = Field(
+        type=np.ndarray, dtype=np.uint64, ndim=1, description="events ucts timestamp"
+    )
+
+    event_type = Field(
+        type=np.ndarray, dtype=np.uint8, ndim=1, description="trigger event type"
+    )
+
+    event_id = Field(type=np.ndarray, dtype=np.uint32, ndim=1, description="event ids")
+
+    amp_int_per_pix_per_event = Field(
+        type=np.ndarray,
+        dtype=np.float32,
+        ndim=3,
+        description="The amplitude integrated over the window width, per pixel and per event",
+    )
+
+    t_peak_per_pix_per_event = Field(
+        type=np.ndarray,
+        dtype=np.uint8,
+        ndim=3,
+        description="Sample containing the pulse maximum, per pixel and per event",
+    )
+
+    FF_coef = Field(
+        type=np.ndarray,
+        dtype=np.float32,
+        ndim=3,
+        description="The flat field coefficient, per event",
+    )
+
+    # masked_wfs = Field(
+    #    type=np.ndarray, dtype=np.uint64, ndim=4, description="Masked array for amplitude integration"
+    # )
+
+    bad_pixels = Field(
+        type=List,
+        dtype=np.uint16,
+        ndim=1,
+        description="Pixels considered as bad in at least one gain channels",
+    )
--- a/src/nectarchain/data/container/flatfieldContainer.py
+++ b/src/nectarchain/data/container/flatfieldContainer.py
@@ -1,0 +1,5 @@
+import logging
+
+logging.basicConfig(format="%(asctime)s %(name)s %(levelname)s %(message)s")
+log = logging.getLogger(__name__)
+log.handlers = logging.getLogger("__main__").handlers
--- a/src/nectarchain/makers/component/preFlatFieldComponent.py
+++ b/src/nectarchain/makers/component/preFlatFieldComponent.py
@@ -1,0 +1,152 @@
+import logging
+
+logging.basicConfig(format="%(asctime)s %(name)s %(levelname)s %(message)s")
+log = logging.getLogger(__name__)
+log.handlers = logging.getLogger("__main__").handlers
+
+import os
+import pathlib
+
+import matplotlib.pyplot as plt
+import numpy as np
+from ctapipe.containers import EventType, Field
+from ctapipe.coordinates import EngineeringCameraFrame
+from ctapipe.core import Component
+from ctapipe.core.traits import ComponentNameList, Dict, Float, Integer, List, Tuple
+from ctapipe.instrument import CameraGeometry
+from ctapipe.io import HDF5TableReader
+from ctapipe.visualization import CameraDisplay
+from ctapipe_io_nectarcam import constants
+from ctapipe_io_nectarcam.containers import NectarCAMDataContainer
+
+from nectarchain.data.container import (
+    ArrayDataContainer,
+    NectarCAMContainer,
+    TriggerMapContainer,
+)
+from nectarchain.makers import EventsLoopNectarCAMCalibrationTool
+from nectarchain.makers.component import ArrayDataComponent, NectarCAMComponent
+from nectarchain.utils import ComponentUtils
+
+__all__ = ["preFlatFieldComponent"]
+
+
+class preFlatFieldComponent(NectarCAMComponent):
+    window_shift = Integer(
+        default_value=5,
+        help="the time in ns before the peak to integrate charge",
+    ).tag(config=True)
+
+    window_width = Integer(
+        default_value=12,
+        help="the duration of the extraction window in ns",
+    ).tag(config=True)
+    ## --< final window is 14 samples ! >--
+
+    g = Float(
+        default_value=58.0,
+        help="defaut gain value",
+    ).tag(config=True)
+
+    hi_lo_ratio = Float(
+        default_value=13.0,
+        help="defaut gain value",
+    ).tag(config=True)
+
+    ## Simple function to substract the pedestal using the 20 first samples of each trace
+    def substract_pedestal(wfs, window=20):
+        ped_mean = np.mean(wfs[0][:, :, 0:window], axis=2)
+        wfs_pedsub = wfs - np.expand_dims(ped_mean, axis=-1)
+        return wfs_pedsub
+
+    ## Function to make an array that will be used as a mask on the waveform for the calculation of the integrated amplitude of the signal.
+    def make_masked_array(t_peak, window_shift, window_width):
+        masked_wfs = [
+            np.array([np.zeros(constants.N_SAMPLES)] * constants.N_PIXELS)
+        ] * constants.N_GAINS
+        masked_wfs = np.array(masked_wfs)
+
+        t_signal_start = t_peak - window_shift
+        t_signal_stop = t_peak + window_width - window_shift
+
+        for g in range(0, constants.N_GAINS):
+            for i in range(0, constants.N_PIXELS):
+                masked_wfs[g][i][
+                    t_signal_start[0, g, i] : t_signal_stop[0, g, i]
+                ] = True
+                # --< I didn't find a better way to do than using this masked array >--
+
+    return masked_wfs
+
+    def __init__(self, subarray, config=None, parent=None, *args, **kwargs):
+        super().__init__(
+            subarray=subarray, config=config, parent=parent, *args, **kwargs
+        )
+
+        self.__ucts_timestamp = []
+        self.__event_type = []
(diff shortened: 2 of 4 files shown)


jlenain (commented):
Hi @ArmelleJB !
Many thanks for this PR. I'll mark it as draft for the time being, as I understand it is not yet complete.
At some point, in your user script, could you provide an example on how to launch a flatfield computation please ?

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+        t_signal_stop = t_peak + window_width - window_shift
+
+        for g in range(0, constants.N_GAINS):
+            for i in range(0, constants.N_PIXELS):
+                masked_wfs[g][i][
+                    t_signal_start[0, g, i] : t_signal_stop[0, g, i]
+                ] = True
+                # --< I didn't find a better way to do than using this masked array >--
+
+    return masked_wfs
```
jlenain, line 79: This return is outside of its function. Indentation missing ?

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+            self.__amp_int_per_pix_per_event.append(amp_int_per_pix_per_event)
+            mean_amp_cam_per_event = np.mean(amp_int_per_pix_per_event, axis=-1)
+
+            # get efficiency and flat field coefficient
+            gain = [self.g, self.g / self.hi_lo_ratio]
+
+            eff = np.divide(
+                (amp_int_per_pix_per_event[:] / (np.expand_dims(gain[:], axis=-1))),
+                np.expand_dims((mean_amp_cam_per_event[:] / gain[:]), axis=-1),
+            )  # efficacité relative
```
jlenain, line 124: I understand this is the meat of the FF computation, but I have a hard time understanding these 3 lines.

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+        if event.trigger.event_type.value == EventType.FLATFIELD.value:
+            # print("event :", (self.__event_id, self.__event_type))
+            self.__event_id.append(np.uint32(event.index.event_id))
+            self.__event_type.append(event.trigger.event_type.value)
+            self.__ucts_timestamp.append(event.nectarcam.tel[0].evt.ucts_timestamp)
+
+            wfs.append(event.r0.tel[0].waveform)  # not saved
+
+            # substract pedestal using the mean of the 20 first samples
+            wfs_pedsub = substract_pedestal(wfs, 20)
```
jlenain, line 105: self missing: `self.subtract_pedestal` ?

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+    ).tag(config=True)
+
+    ## Simple function to substract the pedestal using the 20 first samples of each trace
+    def substract_pedestal(wfs, window=20):
+        ped_mean = np.mean(wfs[0][:, :, 0:window], axis=2)
+        wfs_pedsub = wfs - np.expand_dims(ped_mean, axis=-1)
+        return wfs_pedsub
+
+    ## Function to make an array that will be used as a mask on the waveform for the calculation of the integrated amplitude of the signal.
+    def make_masked_array(t_peak, window_shift, window_width):
```
jlenain, line 63: Is it a static method ? Otherwise, I guess `self` is missing in the signature.

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+        help="defaut gain value",
+    ).tag(config=True)
+
+    hi_lo_ratio = Float(
+        default_value=13.0,
+        help="defaut gain value",
+    ).tag(config=True)
+
+    ## Simple function to substract the pedestal using the 20 first samples of each trace
+    def substract_pedestal(wfs, window=20):
```
jlenain, line 57: Is it a static method ? Otherwise, I guess `self` is missing in the signature.

src/nectarchain/data/container/flatfieldContainer.py
```diff
@@ -0,0 +1,85 @@
...
+    # masked_wfs = Field(
+    #    type=np.ndarray, dtype=np.uint64, ndim=4, description="Masked array for amplitude integration"
+    # )
+
+    bad_pixels = Field(
+        type=List,
+        dtype=np.uint16,
+        ndim=1,
+        description="Pixels considered as bad in at least one gain channels",
+    )
```
jlenain, line 85: I am wondering whether we should not try to use the `FlatFieldContainer` from `ctapipe` here:
https://ctapipe.readthedocs.io/en/v0.19.3/api/ctapipe.containers.FlatFieldContainer.html

ArmelleJB: Yes I guess that it's an option, if it's more convenient.

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+
+from nectarchain.data.container import (
+    ArrayDataContainer,
+    NectarCAMContainer,
+    TriggerMapContainer,
+)
+from nectarchain.makers import EventsLoopNectarCAMCalibrationTool
+from nectarchain.makers.component import ArrayDataComponent, NectarCAMComponent
+from nectarchain.utils import ComponentUtils
+
```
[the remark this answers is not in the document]
ArmelleJB: True - I cleaned it.

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: Right now yes - but I don't know if it's the more clever thing to do

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: Right now yes

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+            self.__ucts_timestamp.append(event.nectarcam.tel[0].evt.ucts_timestamp)
+
+            wfs.append(event.r0.tel[0].waveform)  # not saved
+
+            # substract pedestal using the mean of the 20 first samples
+            wfs_pedsub = substract_pedestal(wfs, 20)
+
+            # get the masked array for integration window
+            t_peak = np.argmax(wfs_pedsub, axis=3)
+            masked_wfs = make_masked_array(t_peak, self.window_shift, self.window_width)
```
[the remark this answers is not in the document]
ArmelleJB: make_masked_array is a static function (defined before the init) - but maybe it's not the proper way to do it.

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: subtract_pedestal is a simple static function (defined before the init)

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: Absolutely - corrected.

src/nectarchain/makers/component/preFlatFieldComponent.py
```diff
@@ -0,0 +1,152 @@
...
+        wfs_pedsub = wfs - np.expand_dims(ped_mean, axis=-1)
+        return wfs_pedsub
+
+    ## Function to make an array that will be used as a mask on the waveform for the calculation of the integrated amplitude of the signal.
+    def make_masked_array(t_peak, window_shift, window_width):
+        masked_wfs = [
+            np.array([np.zeros(constants.N_SAMPLES)] * constants.N_PIXELS)
+        ] * constants.N_GAINS
+        masked_wfs = np.array(masked_wfs)
+
```
vmarandon, line 68: Maybe the line could be simplified by : 
`masked_wfs = np.zeros( shape=(constants.N_GAINS,constants.N_PIXELS,constants.N_SAMPLES))`

You want to create a mask but by default of bumpy array the type is float.
I would add `dtype=bool` to the constructor above to truly have a boolean mask.

src/nectarchain/makers/component/preFlatFieldComponent.py
vmarandon: Regarding your comment about a better way, I think something like those lines should work:

```
## Assume that t_peak is of shape (n_gain, n_pixels)
slice_times = np.expand_dims( np.arange(N_SAMPLES), axis=(0,1) )
t_peaks = np.expand_dims(t_peak,axis=-1)
t_mins = t_peaks - window_shift
t_maxs = t_peaks - window_shift + window_width
masked_wfs = (slice_times>=t_mins) & (slice_times<t_maxs)
```

ArmelleJB: My note was to point out that I didn't find a better way than using a mask array. But there are certainly better ways of coding it ;-)

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: OK - changed.

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: It's probably more explicite and cleaner that way :
            amp_int_per_pix_per_event_pe = amp_int_per_pix_per_event[:]/(np.expand_dims(gain[:], axis=-1))
            mean_amp_cam_per_event_pe = np.mean(amp_int_per_pix_per_event_pe, axis=-1)
            
            eff = np.divide(amp_int_per_pix_per_event_pe, np.expand_dims(mean_amp_cam_per_event_pe, axis=-1))

src/nectarchain/makers/component/preFlatFieldComponent.py
vmarandon: Oups! Sorry... 😅

jlenain: I guess using a masked array should be just fine. But what about @vmarandon's remark about edge effects ? Should we account for them ?

src/nectarchain/makers/component/preFlatFieldComponent.py
jlenain: OK, I see you explicitly declared it as a static method now, all fine!

src/nectarchain/makers/component/preFlatFieldComponent.py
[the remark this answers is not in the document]
jlenain: Same here, explicitly stated as a static method.

src/nectarchain/data/container/flatfieldContainer.py
jlenain: I was just wondering, and am really not sure at the moment if it is the best option: the `FlatFieldContainer` in `ctapipe` has other attributes, with almost none in common with those you implemented.

tibaldo: Thank you very much for all this work @ArmelleJB! I think it would be good for record keeping to write down somewhere (in this thread?) what are the different intents/attributes of the `FlatFieldContainer` in `ctapipe` and the one you are proposing to add to nectarchain.

src/nectarchain/makers/component/preFlatFieldComponent.py
ArmelleJB: Yes, here I assume that the pulse is in the middle of the trace, or let say not on the edge. I have to think about it.

jlenain:
Hi!

The PR #163 has been merged, which results in a couple of small conflicts in this one that I'll try to resolve.

jlenain:
Note that I had to rebase this branch from `main`, but a good point is that the CI/Lint job now passes !

src/nectarchain/data/container/flatfield_container.py
jlenain, line 1: Could you please rename this script file and remove `_test` from its name? Otherwise, `pytest` (used in CI for the unit tests) will presumably clash.

jlenain: Done

src/nectarchain/makers/component/preflatfield_component.py
```diff
@@ -0,0 +1,183 @@
...
+            # get the masked array for integration window
+            t_peak = np.argmax(wfs_pedsub, axis=3)
+            masked_wfs = self.make_masked_array(
+                t_peak, self.window_shift, self.window_width
+            )
+            # --< I didn't find a better way to do than using this masked array >--
+
+            # get integrated amplitude and mean amplitude over all pixels per event
+            amp_int_per_pix_per_event = np.sum(
+                wfs_pedsub[0], axis=2, where=masked_wfs.astype("bool")
```
jlenain, line 95: I guess "0" here refers to the high gain channel, is that right ? In such a case, it would be better to use `ctapipe_io_nectarcam.constant.HIGH_GAIN` instead, in order to avoid hardcoding stuff.

src/nectarchain/makers/component/preflatfield_component.py
```diff
@@ -0,0 +1,183 @@
...
+            self.__event_type.append(event.trigger.event_type.value)
+            self.__ucts_timestamp.append(event.nectarcam.tel[0].evt.ucts_timestamp)
+
+            wfs.append(event.r0.tel[0].waveform)  # not saved
+
+            # subtract pedestal using the mean of the 20 first samples
+            wfs_pedsub = self.subtract_pedestal(wfs, 20)
+
+            # get the masked array for integration window
+            t_peak = np.argmax(wfs_pedsub, axis=3)
```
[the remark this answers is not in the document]
jlenain: No worry at all! Actually, I am never clear whether we should use a global or local charge integrator for the flat field estimation.

To be checked, but it may even be that the `ctapipe` charge extractors have a correction option when the pulses are on the edges of the time window, as we discussed the last couple of days.

src/nectarchain/makers/component/preflatfield_component.py
ArmelleJB: Ah, actually no, this zero was calling the first event (and there is always only one event in wfs and wfs_sum) - I fixed that.

src/nectarchain/makers/component/preflatfield_component.py
```diff
@@ -67,42 +71,44 @@ def __init__(self, subarray, config=None, parent=None, *args, **kwargs):
...
+
+            # mask bad pixels
+            self.__bad_pixels.append(self.bad_pix)
+            masked_wfs[:, self.bad_pix, :] = False
 
             # get integrated amplitude and mean amplitude over all pixels per event
-            amp_int_per_pix_per_event = np.sum(
-                wfs_pedsub[0], axis=2, where=masked_wfs.astype("bool")
-            )
+            amp_int_per_pix_per_event = np.sum(wfs_pedsub, axis=-1, where=masked_wfs)
```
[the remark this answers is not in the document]
ArmelleJB: Yes we can ! I removed it because I now define the mask array as boolean, but it doesn't hurt to keep it here.

jlenain (approved):
Many thanks again, @ArmelleJB !

PR #163, opened 2025-01-09
Unit tests for makers, container and component modules (part 1/X)

guillaumegrolleron:
This PR is the first part of the implementation of the unit tests for : 
```
nectarchain.data.container
nectarchain.makers (ChargesMakers, WaveformsMakers)
```
It also includes modification of the core containers class to make more robust the I/O methods. Thus the pedestal workflow has been changed to fully use the `Necrtachain` containers (instead of the `pytable` module). 

The LightNectarcamEventSource has been removed because you are now using the implementation of `ctapipe_io_nectarcam`

There are also improvement of user scripts for waveforms, charges and photostat computation

PR #168, opened 2025-01-17
Unit test (part 2)

guillaumegrolleron:
This PR is the follow-up of the PR #163.

Unit tests are implemented for all the `component` module and SPE calibration `makers`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,17 @@ jobs:
         shell: bash -leo pipefail {0}
 
     steps:
+      - name: Install dependencies for Qt 5 on Ubuntu
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get update
+          sudo apt-get install freeglut3-dev
+
+      - name: Install dependencies for Qt 5 on macOS
+        if: matrix.os == 'macos-14'
+        run: |
+          brew install freeglut
+
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
@@ -108,17 +119,6 @@ jobs:
           sudo apt-get install '^libxcb.*-dev' libx11-xcb-dev libglu1-mesa-dev libxrender-dev libxi-dev libxkbcommon-dev libxkbcommon-x11-dev
           # start xvfb in the background
           sudo /usr/bin/Xvfb $DISPLAY -screen 0 1280x1024x24 &
-
-      - name: Install dependencies for Qt 5 on Ubuntu
-        if: matrix.os == 'ubuntu-latest'
-        run: |
-          sudo apt-get update
-          sudo apt-get install freeglut3-dev
-
-      - name: Install dependencies for Qt 5 on macOS
-        if: matrix.os == 'macos-14'
-        run: |
-          brew install freeglut
 
       - name: Install dependencies
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
       - name: Install dependencies for Qt 5 on macOS
         if: matrix.os == 'macos-14'
         run: |
-          brew install qt5
+          brew install freeglut
 
       - name: Install dependencies
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,9 +120,8 @@ jobs:
         run: |
           brew install freeglut
           brew install pyqt@6
-          echo "DYLD_FRAMEWORK_PATH=$(brew --prefix qt)/lib" >> $GITHUB_ENV
+          #echo "DYLD_FRAMEWORK_PATH=$(brew --prefix qt)/lib" >> $GITHUB_ENV
           #echo "QT_QPA_PLATFORM=offscreen" >> $GITHUB_ENV
-          
 
       - name: Install dependencies
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,13 +115,13 @@ jobs:
           # start xvfb in the background
           sudo /usr/bin/Xvfb $DISPLAY -screen 0 1280x1024x24 &
       
-      - name: Instal QT and setup environment on macOS
-        if: matrix.os == 'macos-14'
-        run: |
-          brew install freeglut
-          brew install pyqt@6
-          echo "DYLD_FRAMEWORK_PATH=$(brew --prefix qt)/lib" >> $GITHUB_ENV
-          echo "QT_QPA_PLATFORM=offscreen" >> $GITHUB_ENV
+      #- name: Instal QT and setup environment on macOS
+      #  if: matrix.os == 'macos-14'
+      #  run: |
+          #brew install freeglut
+          #brew install pyqt@6
+          #echo "DYLD_FRAMEWORK_PATH=$(brew --prefix qt)/lib" >> $GITHUB_ENV
+          #echo "QT_QPA_PLATFORM=offscreen" >> $GITHUB_ENV
           
 
       - name: Install dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
           check-latest: true
 
       - name: Install libglib2
-        run: sudo apt-get install -y libglib2.0-0
+        run: sudo apt install -y libglib2.0-0
 
       - name: Install dependencies
         env:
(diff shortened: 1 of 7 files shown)


guillaumegrolleron:
Hi @jlenain , I definitely don't know to fix the issue with pyqt, do you have any idea ?

jlenain:
> Hi @jlenain , I definitely don't know to fix the issue with pyqt, do you have any idea ?

Hi @guillaumegrolleron ,
I need to play around this as well, but as some point during your tests, the CI/tests was successful at least on the macOS platform, and `pyqt` and `pyqtgraph` were successfully installed on Linux as well. Then, the CI was failing on missing `libEGL.so.1` library: see e.g. https://github.com/cta-observatory/nectarchain/actions/runs/12872852602.

jlenain:
I am playing around in a specific branch in my fork (https://github.com/jlenain/nectarchain/tree/refs/heads/dev_test_doc_2), and it seems to boil down to failing runners on `src/nectarchain/makers/component/tests/test_flatfield_spe_component.py::TestFlatFieldCombinedSPEStdNectarCAMComponent::test_finish_multiproc`. I am starting to wonder whether this specific test might too demanding on the GitHub CI platform, see https://github.com/jlenain/nectarchain/actions/runs/12876447806/job/35899468610.

jlenain:
@guillaumegrolleron , with the Xvfb server enabled, tests are now passing on Linux, it seems, even with multiprocessing activated both on the `pytest` side and in `test_flatfield_spe_component`.

I am now struggling to find a way to get `QtCore` available on macOS runners...

guillaumegrolleron:
> @guillaumegrolleron , with the Xvfb server enabled, tests are now passing on Linux, it seems, even with multiprocessing activated both on the `pytest` side and in `test_flatfield_spe_component`.
> 
> I am now struggling to find a way to get `QtCore` available on macOS runners...

Yes now it works with multiprocessing with Xvfb, but after you find a way to fix QT on macos I will add the installation of Xvfb on macOS  as well

jlenain:
Or, since https://github.com/cta-observatory/nectarchain/actions/runs/12834657253/job/35792446232 was passing on macOS, let's try to get back to that configuration for macOS.

guillaumegrolleron:
> OK, I am no macOS expert, and do not know exactly how to get pyqt5 working properly on macOS arm64. We may even hit the problem of having, or not, to use Apple's virtualization layer, Rosetta, to get this done. No idea how to get that up and running on the GitHub CI platform.
> 
> @guillaumegrolleron , on another side, is it possible to avoid the call to `SPEnominalalgorithm.display` in unit tests ?

It is, with the price to reduce the coverage

jlenain:
However, with the last commit, we are back to the point where the CI/tests fails on macOS due to display issues with `pyqtgraph`.

guillaumegrolleron:
Hi @jlenain 

It was a mess but I actually found a solution : 

- for Ubuntu :  

1. the `freeglut3-dev` package have to be installed in the CI
2. the `Xvfb` package to run a X11 server

- For macOS it is more complex : for a strange reason we do not face X11 error, however `PyQt5` is not working well on macOS, then I changed the version to `PyQt6`. The installation is done through `pip`, but, we also need to install with `brew` the `pyqt6` package to fix the dynamic library (like `freeglut3` on Ubuntu) and to manually set the environment variable `DYLD_FRAMEWORK_PATH`. Moreover, we still need to install `pyqt` with `mamba`. It does not give any sense for me but it works...

I'll fix in a upcoming PR the `PyQt5` dependencies in the `TRR_suite`

jlenain:
OK, thanks a lot, @guillaumegrolleron , what a journey! I'll have a look at the rest of the changes, but a first pass already looked good to me. Final review coming ASAP.

.github/workflows/ci.yml
```diff
@@ -132,6 +162,11 @@ jobs:
         with:
           python-version: 3.9
 
+      - name: Install dependencies for Qt 5 on Ubuntu
```
jlenain, line 165: Is it also needed to install Qt 5 "just" to build the doc ?

pyproject.toml
```diff
@@ -27,6 +27,8 @@ dependencies = [
     "scipy==1.11.4",
     "zodb",
     "zeo",
+    "pyqt6",
```
jlenain, line 30: I guess this will install `pyqt` version 6 for `pip` installation method both on Linux and macOS. Is this an issue ?

src/nectarchain/makers/component/tests/test_flatfield_spe_component.py
```diff
@@ -0,0 +1,287 @@
...
+            "nproc": self.NPROC,
+        }
+        assert isinstance(instance.chargesComponent, ChargesComponent)
+        assert instance._chargesContainers is None
+
+    def test_call(self, instance, event):
+        instance(event)
+        assert len(instance.chargesComponent.trigger_list) == 1
+
+    @pytest.mark.skip(reason="test multiproc make GitHub worker be killed")
```
jlenain, line 74: So we indeed need to avoid using multi-threading in GitHub workers, you confirm it, @guillaumegrolleron ?

src/nectarchain/user_scripts/dmousadi/TRR_scripts/src/tools_components.py
```diff
@@ -12,12 +12,13 @@
 from ctapipe_io_nectarcam.containers import NectarCAMDataContainer
 from scipy.interpolate import InterpolatedUnivariateSpline
 from scipy.signal import find_peaks
-from utils import adc_to_pe, argmedian
 
 from nectarchain.data.container import NectarCAMContainer
 from nectarchain.makers import EventsLoopNectarCAMCalibrationTool
 from nectarchain.makers.component import NectarCAMComponent
 
+from .utils import adc_to_pe, argmedian
```
jlenain, line 20: But then, the TRR GUI fails to start with:

```
python gui.py 
Traceback (most recent call last):
  File "/home/jlenain/local/src/python/cta-observatory/nectarchain/src/nectarchain/user_scripts/dmousadi/TRR_scripts/gui.py", line 47, in <module>
    import deadtime
  File "/home/jlenain/local/src/python/cta-observatory/nectarchain/src/nectarchain/user_scripts/dmousadi/TRR_scripts/src/deadtime.py", line 12, in <module>
    from tools_components import DeadtimeTestTool
  File "/home/jlenain/local/src/python/cta-observatory/nectarchain/src/nectarchain/user_scripts/dmousadi/TRR_scripts/src/tools_components.py", line 20, in <module>
    from .utils import adc_to_pe, argmedian
ImportError: attempted relative import with no known parent package
```
In my opnion, better to keep that for another PR where the TRR test suite is integrated into the core code.

src/nectarchain/user_scripts/dmousadi/TRR_scripts/src/pedestal.py
```diff
@@ -10,6 +10,9 @@
 from tools_components import PedestalTool
 from utils import adc_to_pe, pe2photons
 
+from .tools_components import PedestalTool
+from .utils import adc_to_pe, pe2photons, photons2pe
```
jlenain, line 14: Same comment as https://github.com/cta-observatory/nectarchain/pull/168/files#r1925344792

.github/workflows/ci.yml
guillaumegrolleron: Yes otherwise I faced an issue with imports of nectarchain. I think because now pyqt is explicitly imported on top of one submodule

pyproject.toml
guillaumegrolleron: It's not, just the TRR is running with version 5, but we can update the TRR code

src/nectarchain/makers/component/tests/test_flatfield_spe_component.py
guillaumegrolleron: I think it's just a mistake, I forgot to reactivate these tests

src/nectarchain/user_scripts/dmousadi/TRR_scripts/src/tools_components.py
guillaumegrolleron: Yes it seems to be a mistake, I'm working on 3 different branches in the same time

src/nectarchain/makers/component/tests/test_flatfield_spe_component.py
guillaumegrolleron: Ok these tests work well

PR #183, opened 2025-02-17
Flat field

ArmelleJB:
Since in the end only one component is needed, I renamed it FlatFieldComponent (instead of preFlatFieldComponent).
I added the possibility to use LocalPeakWindowSum and GlobalPeakWindowSum as charge extractor.


jlenain (approved):
Thanks a lot, @ArmelleJB ! Looks good to me.

PR #189, opened 2025-04-08
Fetch files on DIRAC both from /vo.cta.in2p3.fr and from /ctao base directories

jlenain:
Following the recent change of VO name from `vo.cta.in2p3.fr` to `ctao` on DIRAC for the data management system, this PR accommodate for both base paths to automatically fetch files from DIRAC for NectarCAM data.

This is heavily inspired from the CTADIRAC script `cta-prod-get-file` from https://gitlab.cta-observatory.org/cta-computing/dpps/workload/CTADIRAC/-/blob/main/src/CTADIRAC/Core/scripts/cta_prod_get_file.py?ref_type=heads

### DIFF
--- a/src/nectarchain/data/management.py
+++ b/src/nectarchain/data/management.py
@@ -62,6 +62,13 @@ class DataManagement:
                         run_number, basepath="/ctao/nectarcam"
                     )
                 )
+                if not lfns:
+                    not_found = FileNotFoundError(
+                        f"Could not find run {run_number} on DIRAC (neither under "
+                        f"/ctao/nectarcam nor /vo.cta.in2p3.fr/nectarcam)"
+                    )
+                    log.error(not_found, exc_info=True)
+                    raise not_found
                 DataManagement.getRunFromDIRAC(lfns)
                 list = glob.glob(
                     basepath + "**/*" + str(run_number) + "*.fits.fz", recursive=True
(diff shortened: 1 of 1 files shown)


guillaumegrolleron:
Hi @jlenain,
Many thanks for this PR. Everything seems to work smoothly on my side, concerning `nectarchain`, the data locations on the grid are perfectly found. However `dirac` fails to download the data but my proxy seems to be fine...

src/nectarchain/data/management.py
```diff
@@ -54,7 +54,14 @@ def findrun(run_number: int, search_on_GRID=True) -> Tuple[Path, List[Path]]:
...
+                    run_number, basepath="/vo.cta.in2p3.fr/nectarcam"
+                )
+                lfns.extend(
+                    DataManagement.get_GRID_location(
+                        run_number, basepath="/ctao/nectarcam"
+                    )
+                )
                 DataManagement.getRunFromDIRAC(lfns)
                 list = glob.glob(
                     basepath + "**/*" + str(run_number) + "*.fits.fz", recursive=True
```
guillaumegrolleron, line 67: Maybe we could raise here an exception if files are not found, I just realize that it's missing

guillaumegrolleron (approved):

guillaumegrolleron:
> > Hi @jlenain, Many thanks for this PR. Everything seems to work smoothly on my side, concerning `nectarchain`, the data locations on the grid are perfectly found. However `dirac` fails to download the data but my proxy seems to be fine...
> 
> Hi @guillaumegrolleron , many thanks for the check! What error do you get exactly with the download ?

I don't have any that my problem.. I will try to investigate

PR #191, opened 2025-05-21
Upgrade to ctapipe 0.24

jlenain:


jlenain:
> Hi @jlenain, I think to have fixed the problem in the unit tests. The point was related to the PR [cta-observatory/ctapipe#2529](https://github.com/cta-observatory/ctapipe/pull/2529), where `ctapipe` extractors can now deal with waveforms with shape (n_channel, n_pixels, n_samples). The output in `DL1CameraContainer` is then quite different. The modifications in `nectarchain` are small because of the transparency use of extractors

Hi @guillaumegrolleron ,
Many thanks for that !
I also had to ensure the member `pixel_mask` in `PedestalEstimationComponent` was of the expected type to fix the remaining unit tests.
I am still not sure whether we need to enforce the shape of waveforms to (n_gain, n_pixel, n_sample) in `ctapipe_io_nectarcam` for all cases (see https://github.com/cta-observatory/ctapipe_io_nectarcam/pull/70/files#diff-c1dfa027cc2f2c588325ed4878a39479d6483636e83523c960496fbf1cbc62adL1190 and https://github.com/cta-observatory/ctapipe_io_nectarcam/pull/70/files#diff-c1dfa027cc2f2c588325ed4878a39479d6483636e83523c960496fbf1cbc62adL1197) or not. @vmarandon , would you have an idea ?

Now, the CI workflow still fails on macOS, but this related to another issue to be fixed (proper installation of `protozfits`).

src/nectarchain/makers/component/charges_component.py
```diff
@@ -568,44 +566,42 @@ def compute_charges(
...
-            return ChargesContainer.fields["charges_lg"].dtype.type(
-                out[0]
-            ), ChargesContainer.fields["peak_lg"].dtype.type(out[1])
+            index = 1
+            gain_label = "lg"
         else:
             raise ArgumentError(
                 None, f"channel must be {constants.LOW_GAIN} or {constants.HIGH_GAIN}"
             )
+        return ChargesContainer.fields[f"charges_{gain_label}"].dtype.type(
```
guillaumegrolleron, line 602: Here we are computing both for low and high gain channel de charge, however the function return only one of these quantities. It's more elegant, but I'm wondering what is lost in term of performance. (Actually the charge extraction is `numba` based so I guess it is negligible).

guillaumegrolleron (changes requested):

environment.yml
```diff
@@ -4,23 +4,25 @@ channels:
...
+  - scipy
   - sphinx
   - sphinx-automodapi
   - pydata-sphinx-theme
   - lmfit # needed into TRR
   - h5py # needed into TRR (should be removed to use I/O methods of containers)
   - pyqt
+  - protozfits
   - pip:
+    - git+https://github.com/jlenain/ctapipe_io_nectarcam.git@ctapipe-0.24-bis  # TODO: should be removed when merging PR
```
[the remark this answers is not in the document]
jlenain: Yes, it should absolutely be modified before merging this PR.

src/nectarchain/makers/component/charges_component.py
jlenain: Since this method has a mandatory argument for the channel, I'd say it may be more efficient to only compute the charge in the requested channel.

guillaumegrolleron: Hi @jlenain, yes, I think it's perfect like this, and it avoids the loop over waveforms

guillaumegrolleron (approved):
Hi @jlenain, thanks a lot for your update, I approve the PR

PR #203, opened 2025-07-02
DQM optimization

jlenain:
This PR proposes to clean, streamline and optimize the DQM code. At the moment, I benchmarked a speed-up of a factor 3 on the processing time.


jlenain:
In the end, the speed-up factor is about 4-5, and memory consumption is reduced by a factor ~10 on large runs.

hashkar (approved):

PR #206, opened 2025-08-12
update example scripts with use of arguments

ArmelleJB:


src/nectarchain/user_scripts/ajardinb/plot_flatfield_output.py
```diff
@@ -1,12 +1,49 @@
...
-# Define the global environment variable NECTARCAMDATA (folder where are the runs)
-run_number = 4940
-os.environ["NECTARCAMDATA"] = "./20231222/FlatFieldTests"
+logging.basicConfig(format="%(asctime)s %(name)s %(levelname)s %(message)s")
+log = logging.getLogger(__name__)
+log.handlers = logging.getLogger("__main__").handlers
+
+# Option and argument parser
+parser = argparse.ArgumentParser(
+    description="Give run number, run location (if no access to DIRAC) and output location",
```
vmarandon, line 19: Perhaps you could change the description as the second argument is not the run location and there is no third argument.

src/nectarchain/user_scripts/ajardinb/flatfield_example.py
```diff
@@ -1,13 +1,47 @@
...
 
-# Define the global environment variable NECTARCAMDATA (folder where are the runs)
-os.environ["NECTARCAMDATA"] = "./20231222"
+logging.basicConfig(format="%(asctime)s %(name)s %(levelname)s %(message)s")
+log = logging.getLogger(__name__)
+log.handlers = logging.getLogger("__main__").handlers
+
+# Option and argument parser
+parser = argparse.ArgumentParser(
+    description="Give run number, run location (if no access to DIRAC) and output location",
```
vmarandon, line 20: Perhaps you could change the description as there is no third argument.
The "if no access to DIRAC" is maybe not needed: If I understood correctly, if the run is not in the path pointed by "args.path" then it will try to retrieve it from DIRAC and put it there.

vmarandon (approved):
It looks good to me.
I only have minor comments, that I would likely not have done for "real" user scripts, but here, they are examples to use your Tool.

src/nectarchain/user_scripts/ajardinb/flatfield_example.py
ArmelleJB: Done.

src/nectarchain/user_scripts/ajardinb/plot_flatfield_output.py
ArmelleJB: The second argument is the path to the runs folder. There was a third argument that I removed and forgot to update the description.

vmarandon: Did you forgot to add this change to the recent commit ? I don't see the modification.

ArmelleJB: Yes, apparently I committed only one file. Now it should be there.

vmarandon: OK. Looks good now ! Thanks !

PR #207, opened 2025-08-19
Addition of CalibrationWriterNectarCAM tool for writing Category A calibration files

pcorcam:
First attempt at creating a Category A calibration file, inspired by [`lstcam_calib.tools.create_calibration_file`](https://gitlab.cta-observatory.org/cta-array-elements/lst/analysis/lstcam_calib/-/blob/main/src/lstcam_calib/tools/create_calibration_file.py). For now, this `CalibrationWriterNectarCAM` tool works as follows:

1. Take the `.h5` output files from a `NectarCAMCalibrationTool`, which contain results of pedestal, etc., in a `NectarCAMPedestalContainer`, etc.
2. Copy the results in relevant `ctapipe` containers that are used to produce the Category A calibration file. In some cases, this requires some further computations. For example, in `nectarchain`, the `NectarCAMPedestalContainer` contains the average pedestal per pixel _per sample_, while in `ctapipe`, the `PedestalContainer` expects the pedestal per pixel averaged over all samples.
3. Write the Category A calibration file in either `.h5`, `.fits`, or `.fits.gz` formats, as done in `lstcam_calib`.

An usage example of this tool has been added to `user_scripts/pcorrea/example_create_catA_calibration_file.py`.

I chose this way of implementing this tool so I don't have to make any changes to the already existing `nectarchain` tools at this stage.  I also opted to _not_ depend on `lstcam_calib` in `nectarchain` here, since `lstcam_calib` uses their custom `EventSource` in their `CalibrationWriter` tool, which also performs the actual computation of the calibration coefficients for a specific run.

However, this version of the tool is a bit cumbersome since you just copy information from `NectarCAMContainers` to standard `ctapipe` containers. Probably there is a better way to deal with this, but this makes more sense to implement after creating a pipeline that enchains all the `NectarCAMCalibrationTools`.

### DIFF
--- a/src/nectarchain/data/container/core.py
+++ b/src/nectarchain/data/container/core.py
@@ -54,7 +54,9 @@ class NectarCAMContainer(Container):
     """
 
     @staticmethod
-    def _container_from_hdf5(path, container_class, index_component=0):
+    def _container_from_hdf5(
+        path, container_class, index_component=0, group_name="data"
+    ):
         """Static method to read a container from an HDF5 file.
 
         Parameters:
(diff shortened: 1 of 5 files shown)


tibaldo:
Hi @pcorcam, thank you for doing this! I'm a bit confused about the pedestal subtraction. I think that for the NectarCAM we want a correction per sample and not averaged over all samples for each pixel?

PR #221, opened 2025-10-08
Flat Field coefficients computation + the rise and decay times

anmikhno:
The flat field coefficients are computed using the Gaussian fit, on top pf that the outliers are removed to improve the fit. The script generated pdf with the plots and a table with flat-field coefficients computed per pixel using the fit (model method) and flat field coefficients computed using the simple formula (n_pe_per_pix/mean(n_pe_over_camera))

Cleaned branch.

### DIFF
--- a/src/nectarchain/user_scripts/amikhno/flatfield_source_characterization.py
+++ b/src/nectarchain/user_scripts/amikhno/flatfield_source_characterization.py
@@ -589,7 +589,7 @@ def error_propagation_compute(data, minuit_resulting, camera, pdf):
         binned_y * 1.02,
         facecolor="blue",
         alpha=0.5,
-        label="2% requirenment",
+        label="2% requirement",
         zorder=1,
     )
 


src/nectarchain/user_scripts/amikhno/FF_calibration_run6672.dat
jlenain: Please, let's not have data files in the repository. I think I commented the same for PR #239 and PR #251.

src/nectarchain/user_scripts/ajardinb/flatfield_example.py
jlenain: Please do not change other's scripts.

src/nectarchain/makers/calibration/gain/photostat_makers.py
```diff
@@ -135,7 +135,7 @@ def start(
                     f" with extraction"
                     f" method FullWaveformSum,\n reload charges from event loop"
                 )
-
+            print("We are HERE?")
```
jlenain: Please let's not use `print` but `logging` instead. This one was obviously for debugging purposes and can be removed.

jlenain:
@anmikhno , could you please also test that this works for runs taken with NectarCAM 2 ?

src/nectarchain/user_scripts/amikhno/analysis_photostat_simple.py
```diff
@@ -0,0 +1,1000 @@
...
+    formatter = logging.Formatter(
+        "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+    )
+    handler.setFormatter(formatter)
+    if not log.handlers:
+        log.addHandler(handler)
+
+    # --- Assign other variables ---
+    run_number = args.FF_run_number
+    run_path = args.run_path + f"/runs/NectarCAM.Run{run_number}.0000.fits.fz"
```
jlenain: This naming scheme is not compliant with run numbers below 1000.

jlenain:
Affected by #262

jlenain (approved):
Hi @anmikhno !

Thanks a lot for all the work !
I think this PR is now ready to be merged, if you have nothing else to push into it. We could then move ahead and integrate the central parts into the `trr_test_suite` sub-module.

Please let me know.

Issue #224, opened 2025-10-10
Add `display` as a traitlet in `SPEalgorithm` component


pcorcam:
**Please describe the use case that requires this feature.**
Needed to run (subclasses of) the `FlatFieldSPENominalNectarCAMCalibrationTool` when defining the option to display plots using `traitlets.config`.

**Describe the solution you'd like**
Right now, `display` is passed as a `kwarg` to the `SPEalgorithm` component; see also `user_scripts/ggrolleron/gain_SPEfit_computation.py`. I suggest to define `display` as a configurable trait of the `SPEalgorithm` component. That way, we can also pass this option as e.g. `config["FlatFieldSPENominalNectarCAMCalibrationTool"].display = False`. 

**Additional context**
Note that `SPEalgorithm` already contains a method with the name `display`, so need to be careful with names.

guillaumegrolleron:
Hi @jlenain and @pcorcam, 
I definitely agree, it's a good improvement to make the display configurable with a trait !

jlenain:
Thanks.

To avoid shadowing the method `SPEnominalalgorithm.display`, I'd propose to name this field `display_toggle`.

jlenain:
@pcorcam , would #228 work ?

pcorcam:
@jlenain, yes #228 looks perfect! Thanks!

PR #228, opened 2025-10-15
Change display to a trait

jlenain:
This PR proposes to change `display` in `SPEalgorithm` to a `trait`, which should hopefully answer to #224.


src/nectarchain/makers/component/spe/spe_algorithm.py
pcorcam: ### Lines 318 and 386
There is some residual `display` keyword-argument syntax in `SPEalgorithm._get_mean_gaussian_fit`. I suggest the following change (note that both lines are the same):
`if log.getEffectiveLevel() == logging.DEBUG and kwargs.get("display", False):` -> `if log.getEffectiveLevel() == logging.DEBUG and self.display_toggle:`

pcorcam: This looks great, thanks! I will approve the merge request right away :)

PR #243, opened 2025-12-10
Bug fix in pedestal subtraction within `ChargesComponent`

pcorcam:
Small bug fix in pedestal subtraction of waveform within the `ChargesComponent`. 

Before the fix, the subtraction would work if `wfs_hg_tmp` would e.g. be of type `np.float32` (see e.g. run 6954). However, it would _not_ work correctly if `wfs_hg_tmp` would be of type `np.uint16` (see e.g. run 3414). Negative values (where waveform - pedestal < 0) would erroneously be associated to high uint16 values.

This is fixed by:

1.  Not requiring `self.__pedestal_hg.astype(wfs_hg_tmp.dype)` in the subtraction. That way `wfs_hg_tmp` is converted into a float.
2. Clip negative values of the pedestal-subtracted waveform at 0. That way we ensure that we can never have a negative charge after pedestal subtraction, which would again lead to high `uint16` values when using `ChargesComponent.charges_hg`.

Note that in all of the above, we can replace `hg` with `lg`.

### DIFF
--- a/src/nectarchain/makers/tests/test_charges_makers.py
+++ b/src/nectarchain/makers/tests/test_charges_makers.py
@@ -75,8 +75,8 @@ class TestChargesNectarCAMCalibrationTool:
         assert output.charges_lg.shape == (nevents, n_pixels)
         assert isinstance(output.charges_hg, np.ndarray)
         assert isinstance(output.charges_lg, np.ndarray)
-        assert output.charges_hg.dtype == np.uint16
-        assert output.charges_lg.dtype == np.uint16
+        assert output.charges_hg.dtype == np.int32
+        assert output.charges_lg.dtype == np.int32
         assert output.peak_hg.shape == (nevents, n_pixels)
         assert output.peak_lg.shape == (nevents, n_pixels)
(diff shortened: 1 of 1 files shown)


jlenain:
@guillaumegrolleron , @frnbrun , @tibaldo , @vmarandon , does anyone of you recall why charges should be stored as `np.uint16` ?

jlenain:
> Hi @jlenain, @pcorcam
> 
>     * I don't know why uint16 was chosen.
> 
>     * Storing raw charges (without pedestal subtraction) as integers, makes sense.
> 
>     * I agree that negative values in the pedestal subtracted waveforms should not be clipped.
> 
>     * As far as I can think, pedestal-subtracted charge logically should be stored in a way that allows for negative values unless this poses problems afterwards in the data analysis chain.

Hi @tibaldo ,

Thanks a lot for your reply. That is indeed what I was thinking as well. @guillaumegrolleron , would you anticipate any side effect if we were to store charges as signed int (`np.int16`) instead of unsigned ones ?

jlenain (approved):

PR #239, opened 2025-11-24
TRR Analysis script updates

medha1996c:
Scripts for ChargeResolution using FF + NSB runs are added. Module added to reject bad_pixels, use per pixel per temperature gain (from SPE results by Halim) and FF coefficients using Anastasiia's script output.

PR #244, opened 2025-12-15
Removal of code duplication for charge computation in `trr_test_suite`

pcorcam:
Removes code duplication in the `trr_test_suite` regarding the charge computation. More specifically:

- `ChargeContainer` and `ChargeComp` are removed from `trr_test_suite/tools_components.py`;
- The standard `ChargesComponent` (`makers/components/charges_components.py`) is now the component in the `LinearityTestTool` (`trr_test_suite/tools_components.py`);
- The script `trr_test_suite/linearity.py` has been adapted accordingly - pedestals are precomputed with the standard `PedestalNectarCAMCalibrationTool` and its output is fed to the `ChargesComponent` in the `LinearityTestTool`.

A crosscheck was performed to test the compatibility of both charge-extraction methods. The parameters:
- `run_number = 6954`
- `max_events = 500`
- `window_width = 16`
- `window_shift = 6`
- For standard `ChargesComponent`:
  - `method=LocalPeakWindowSum`
  - pedestal is computed with `PedestalNectarCAMCalibrationTool` with parameters:
    - `max_events = 12000`
    - `events_per_slice = 5000`
    - `method = FullWaveFormSum`
    - `filter_method = None`

We find good agreement between the two charge-extraction methods (see figures below); the relative difference is on average $\mu = 0.0045$% with a standard deviation $\sigma = 0.5335$%.

image
image

Note that there is still quite some code duplication regarding charge computation in `trr_test_suite/tools_components.py`, notably in `ToMComp`. This will be tackled in a future PR.

### DIFF
--- a/src/nectarchain/trr_test_suite/tools_components.py
+++ b/src/nectarchain/trr_test_suite/tools_components.py
@@ -16,7 +16,7 @@ from scipy.signal import find_peaks
 from nectarchain.data.container import NectarCAMContainer
 from nectarchain.makers import EventsLoopNectarCAMCalibrationTool
 from nectarchain.makers.component import NectarCAMComponent
-from nectarchain.trr_test_suite.utils import adc_to_pe
+from nectarchain.utils.constants import GAIN_DEFAULT
 
 
 # overriding so we can have maxevents in the path
@@ -79,8 +79,8 @@ class LinearityTestTool(EventsLoopNectarCAMCalibrationTool):
         charge_lg = charge_container["charges_lg"]
         npixels = charge_container["npixels"]
 
-        charge_pe_hg = np.array(charge_hg) / adc_to_pe
-        charge_pe_lg = np.array(charge_lg) / adc_to_pe
+        charge_pe_hg = np.array(charge_hg) / GAIN_DEFAULT
+        charge_pe_lg = np.array(charge_lg) / GAIN_DEFAULT
 
         for channel, charge in enumerate([charge_pe_hg, charge_pe_lg]):
             pix_mean_charge = np.mean(charge, axis=0)  # in pe


jlenain:
Hi @pcorcam ,

Many thanks for this ! There was a conflict with the freshly merged PR #243 . I rebased this PR against `main`.

jlenain:
Thanks a lot also for the comparison _before_/_after_, @pcorcam, I think this is really compelling!

src/nectarchain/trr_test_suite/linearity.py
```diff
@@ -134,13 +136,27 @@ def main():
...
+        pedestal_tool = PedestalNectarCAMCalibrationTool(
+            progress_bar=True,
+            run_number=run,
+            max_events=12000,
+            events_per_slice=5000,
+            log_level=20,
+            output_path=output_dir + f"/pedestal_{run}.h5",
+            overwrite=True,
+            filter_method=None,
+            method="FullWaveformSum",  # charges over entire window
```
jlenain: Hum, I would have a concern here: the runs being processed for the linearity evaluation are typically flat-field runs acquired with different attenuation factors. They are thus typically non-pedestal runs, in the sense that pulsed events should dominate. The pedestal evaluation should thus, to my understanding, use another method than `FullWaveformSum`. 

I think I also raised the same question in the PR #239, see https://github.com/cta-observatory/nectarchain/pull/239#discussion_r2597737657

@tibaldo , could you please advise about that ?

tibaldo: Hi @jlenain. I think the code should work correctly. The runs processed are flat-field with interleaved pedestal events. `PedestalNectarCAMCalibrationTool` uses only events of the type SKYPEDESTAL to estimate the pedestal.

If you are worried about a contamination from the flasher in interleaved pedestal events then I would recommend to use
```
filter_method="ChargeDistributionFilter",
charge_sigma_high_thr=3.0,
```
to exclude the tails of the charge distribution.


The subsequent call to the linearity tool does not use the charge in the output pedestal file but just the mean waveform values, so the integration method should be irrelevant unless you apply filtering as suggested above (which is not the case in the current version of the code with `filter_method=None`).

PR #250, opened 2026-01-19
Flat field

jlenain:
Proposed by @ArmelleJB initially in PR #245 :

- save the ID of the identified bad pixels (instead of their position in the list), 
- fix format issues when using LocalPeakWindowSum or GlobalPeakWindowSum methods
- add a parameter for the number of samples used for the pedestal estimation 
- add simple check on the location of the peak and discard events that are too close from the beginning of the trace, to safely estimate the pedestal

### DIFF
--- a/src/nectarchain/user_scripts/ajardinb/plot_flatfield_output.py
+++ b/src/nectarchain/user_scripts/ajardinb/plot_flatfield_output.py
@@ -79,9 +79,9 @@ for G in [constants.HIGH_GAIN, constants.LOW_GAIN]:
     )
     plt.legend()
     plt.subplots_adjust(left=0.15, right=0.95, top=0.92, bottom=0.12)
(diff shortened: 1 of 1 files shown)


vmarandon:
> Hi @jlenain, are you talking about this line ? `amp_int_per_pix_per_event[self.__bad_pixels_mask] = 0.0` It is not used because of the following line that mask all the bad pixels using `np.ma.array` (and they are masked again after the if block using `np.ma.masked_array`).
> 
> About the question should we mask the bad pixels (ends up as "--" in the final array) or set them as nan, I don't have a preference. Maybe @vmarandon ?

The main question to me is how the masked array is stored in the Container ? 
Is the mask also stored or is it discarded ? If this is discarded but values kept, then better put them to np.nan. (provided that the nan are correctly handled while re-reading the data from disk)

vmarandon (approved):

PR #256, opened 2026-02-06
Implement python logging in Bokeh app to log app logic

alessandromontanari:
Brief summaries of the changes
- logging_config.py: setup_logger for RotatingFileHandler saving in logs/app.log
- app_hooks.py: use logger instead of print and add new log messages
- main.py: use logger instead of print and add new log messages

### DIFF
--- a/src/nectarchain/dqm/bokeh_app/app_hooks.py
+++ b/src/nectarchain/dqm/bokeh_app/app_hooks.py
@@ -18,7 +18,8 @@ from ctapipe.instrument import CameraGeometry
 # ctapipe imports
 from ctapipe.visualization.bokeh import CameraDisplay
 from ctapipe_io_nectarcam import constants
-from logging_config import setup_logger
+
+from .logging_config import setup_logger
 
 base_dir = os.path.abspath(os.path.dirname(__file__))
 labels_path = os.path.join(base_dir, "data", "labels.json")
--- a/src/nectarchain/dqm/bokeh_app/__init__.py
+++ b/src/nectarchain/dqm/bokeh_app/__init__.py
@@ -1,0 +1,5 @@
+from .logging_config import setup_logger
+
+__all__ = [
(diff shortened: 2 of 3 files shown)


src/nectarchain/dqm/bokeh_app/app_hooks.py
```diff
@@ -267,44 +265,21 @@ def make_camera_display(source, parent_key, child_key):
...
-    except KeyError:
+        logger.error(
+            f"Exception '{e}', filled camera plot"
+            + f"{parent_key}, {child_key} with zeros"
+        )
+    except KeyError as e:
         image = np.zeros(shape=constants.N_PIXELS)
         display.image = image
-
-    fig = display.figure
```
jlenain: Hum, it seems this PR removes features introduced in PR #253 . Let me try to rebase properly.

src/nectarchain/dqm/bokeh_app/logging_config.py
```diff
@@ -0,0 +1,52 @@
...
+    log_level : int, optional
+        Logging level (e.g., logging.INFO, logging.DEBUG), by default logging.INFO
+
+    Returns
+    -------
+    logging.Logger
+        Configured logger
+    """
+
+    # FIXME: temporary dir for the Bokeh app logs is
```
jlenain: I would suggest to store the logs in the directory defined by the (optional) `$NECTARCHAIN_LOG` environment variable (see https://nectarchain.readthedocs.io/en/latest/user-guide/env-vars.html#optional).

src/nectarchain/dqm/bokeh_app/app_hooks.py
alessandromontanari: Sorry for that, you are right!

jlenain (approved):
Thanks, @alessandromontanari !

PR #264, opened 2026-02-12
Adding a full nectarchain calibration pipeline to the TRR

hashkar:
This script performs a complete calibration including:
1. Pedestal computation
2. Gain (SPE fit) computation
3. Flatfield computation
4. Charge extraction
5. Calibrated charge computation (pedestal subtraction, gain correction, FF correction)
6. Plotting of all calibration parameters individually and vs temperature

The script is designed to be flexible and configurable via command-line arguments,
allowing users to specify run numbers, processing options, and output directories.
It also includes robust path resolution logic to handle the various output locations
used by the nectarchain tools.

Depending on the mode it can run and plot each entity separatly: pedestals, gain, FF and charge 
or it can run in mode full and compute calibrated charges by doing: charge computation, pedestal subtraction, gain and FF corrections. 
 
Example usage in full mode: 

python nectarcam_full_calibration_analysis.py \
  --mode full \                                          
  --pedestal_runs 7020 7077\     
  --gain_runs 7066 7123\     
  --flatfield_runs 7066 7123\     
  --charge_runs 7020 7077\     
  --temperatures 25 20\   
  --max_events_gain 5000

run python nectarcam_full_calibration_analysis.py -h for more options.


src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,2042 @@
...
+                    logging.exception(f"Error {filepath}: {e}")
+                    if h5file.isopen:
+                        h5file.close()
+                    raise
+
+        raise KeyError(
+            f"Cannot find {kind} data in {filepath}.\n"
+            f"  Top-level groups : {top_groups}\n"
+            f"  All dataset paths: {all_keys}"
+        )
```
pcorcam: There is already a [built-in method](https://github.com/cta-observatory/nectarchain/blob/cfdf0c963db87df2f81a3b1af8d8cb44ca2d2de7/src/nectarchain/data/container/core.py#L89) in `nectarchain` to read the data containers written on h5 output files. Please use that to read the h5 files in order to not introduce more code duplication in `nectarchain`; through the containers you can directly access all the data fields you need. An example would be something like:

```
pedestal_container = NectarCAMPedestalContainer.from_hdf5("/path/to/pedestal.h5")
pedestal_hg = pedestal_container["pedestal_mean_hg"]
```

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,2042 @@
...
+        self.figure_dir = Path(
+            args.figure_dir or os.environ.get("NECTARCHAIN_FIGURES", "/tmp")
+        )
+
+        # Pipeline-managed subdirectories
+        self.pedestal_dir = self.data_dir / "pedestals"
+        self.gain_dir = self.data_dir / "gains"
+        self.flatfield_dir = self.data_dir / "flatfields"
+        self.charge_dir = self.data_dir / "charges"
+        self.calibrated_dir = self.data_dir / "calibrated_charges"
```
pcorcam: You don't use this, please remove

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,2042 @@
...
+
+        tool.initialize()
+        tool.setup()
+        tool.start()
+        tool.finish(return_output_component=True)[0]
+
+        self.flatfield_results[run_number] = output_path_final
+        self.log.info(f"Flatfield saved to {output_path_final}")
+
+        return output_path_final
```
pcorcam: The `FlatFieldNectarCAMCalibrationTool` is not using the pedestal and gain computation that you can calculate in previous steps. Through `FlatFieldComponent`, there are arguments to supply both a `pedestal_file` and a `gain_file` to the `FlatFieldNectarCAMCalibrationTool`. If the files are not found, the tool automatically falls back to default values. This also means you don’t need to apply the two-pass method for the flatfield computation; one pass is sufficient.

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,2042 @@
...
+        if Path(expected_path).exists():
+            return Path(expected_path)
+
+        # 3. Search default dirs (useful when tool already ran previously)
+        filename = Path(expected_path).name
+        found = self._find_file(filename, kind)
+        if found:
+            return found
+
+        return Path(expected_path)
```
pcorcam: This code seems a bit redundant. You can always initialize a tool, and then get the output path directly from `tool.output_path`. And then you can check if e.g. `os.path.exists(tool.output_path)`.

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,2042 @@
...
+        )
+        return gain
+
+    def get_hi_lo_ratio(self, ff_output):
+        """Calculate high gain to low gain ratio"""
+        gain = self.get_gain_from_flatfield(ff_output)
+        hi_lo_ratio = gain[constants.HIGH_GAIN] / gain[constants.LOW_GAIN]
+        return hi_lo_ratio
+
+    def compute_flatfield(self, run_number):
```
pcorcam: Same comment as https://github.com/cta-observatory/nectarchain/pull/264/changes#r2813318496

pcorcam (changes requested):
Hi @hashkar !

Thanks for all the work on this PR. I've taken a global look at the script, and at first sight the functionality looks good to me. I just would not name it a "calibration pipeline" - more like a "observation temperature" TRR use-case script. 

I do have quite a few major comments though. My main concern is that you "re-invent the wheel" a few times when it comes to reading h5 files, finding existing files, and defining output directories. This leads to a lot of (unnecessary) code duplication, which we should avoid. 

In addition, I think we should not use default values as an option for the TRR, but more as a failsafe if, for some reason, a calibration tool did not successfully execute. But we should always try to run a calibration tool if it is needed for a subsequent step in the pipeline.

Just FYI, I have not looked into detail into the production of the plots yet, because I would like to focus on these major concerns first. In any case, I am willing to contribute to further development of this PR if you would find it useful.

Thanks again!

src/nectarchain/trr_test_suite/observation_temperatures.py
hashkar: I do use them. But this relates to another comment and better have a unified output dir for the figures. For now I use $NECTARCHAIN_FIGURES but after that where do you suggest the plots go ?

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,2042 @@
...
+        # pedestals, gains, ff_coef: (2, n_pixels)
+        # Apply calibration: (charge - pedestal) / gain * FF
+        self.log.info("  Applying calibration...")
+        self.log.info(f"    charges shape: {charges.shape}")
+        self.log.info(f"    pedestals shape: {pedestals.shape}")
+        self.log.info(f"    gains shape: {gains.shape}")
+        self.log.info(f"    ff_coef shape: {ff_coef.shape}")
+
+        # Vectorized calibration (much faster than loops)
+        ped_subtracted = charges - pedestals[np.newaxis, :, :]  # Broadcast pedestals
```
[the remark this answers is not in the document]
hashkar: I did not undestand this

src/nectarchain/trr_test_suite/observation_temperatures.py
pcorcam: Sorry I meant only to highlight `self.calibrated_dir`. You don't write e.g. a container with calibrated charges anywhere in the code.

hashkar:
@pcorcam @jlenain  this should solve the issues raised earlier. However I still have a problem and need your helo with it: Using the nectarchain built in methods to read data containers, it does not seem to find the right path. i get for example WARNING - Cannot load pedestal run 6753: group ``/`` does not have a child named ``/data/NectarCAMPedestalContainer_0. I know the pedestals are storted in data_combined and not data. do you ahave a quick fix for this. the def _load_pedestal(path) is in line 213

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -213,51 +213,36 @@ def _read_container(path, container_class, table_name=None, index_component=0):
...
-    for table in all_candidates:
-        try:
-            container = next(_read_container(path, NectarCAMPedestalContainer, table))
-            hg = np.mean(container.pedestal_mean_hg, axis=1)
-            lg = np.mean(container.pedestal_mean_lg, axis=1)
-            return hg, lg
-        except Exception:
-            continue
-    raise RuntimeError(f"Cannot read NectarCAMPedestalContainer from {path}")
+    container = next(NectarCAMPedestalContainer.from_hdf5(str(path)))
```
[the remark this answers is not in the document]
pcorcam: There is a [wrapper function](https://github.com/cta-observatory/nectarchain/blob/25dfb0798f43049a93486e21143da293e315c314/src/nectarchain/utils/utils.py#L156) that could potentially solve your issue:

```
from nectarchain.utils import ContainerUtils
from nectarchain.utils.constants import GROUP_NAMES_PEDESTAL

container = ContainerUtils.get_container_from_hdf5(
    path_to_file,
    NectarCAMPedestalContainer,
    group_names=GROUP_NAMES_PEDESTAL,
)
```

Could you give this a try?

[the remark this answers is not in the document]
pcorcam: Fixed with commits [c9ebd48](https://github.com/cta-observatory/nectarchain/pull/264/commits/c9ebd48b1fbabeba32b807ed438a90fbf2085573), [6df6982](https://github.com/cta-observatory/nectarchain/pull/264/commits/6df69826608d5c49469d399415f45c551bc1072c), [6bfb6ee](https://github.com/cta-observatory/nectarchain/pull/264/commits/6bfb6ee15e1757ff857d28f375cce79b07122646).

hashkar:
@pcorcam I think that adding a treatement of the bad pixels solves most of these problems as you suggested. 
As for the second part, I am not in favor of subtracting the pedestal from the raw charge. The entire goal of this scipt is to get a "calibrated" charge. I am plotting the raw data for diagnostic reasons

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,986 @@
...
+
+    _constants_path = _Path(__file__).resolve().parent.parent / "utils" / "constants.py"
+    _spec = _ilu.spec_from_file_location("constants", _constants_path)
+    _mod = _ilu.module_from_spec(_spec)
+    _spec.loader.exec_module(_mod)
+    FLATFIELD_DEFAULT = _mod.FLATFIELD_DEFAULT
+    GAIN_DEFAULT = _mod.GAIN_DEFAULT
+    GROUP_NAMES_PEDESTAL = _mod.GROUP_NAMES_PEDESTAL
+    HILO_DEFAULT = _mod.HILO_DEFAULT
+    PEDESTAL_DEFAULT = _mod.PEDESTAL_DEFAULT
```
pcorcam, line 34: I don't understand why this exception is necessary... Could you explain please?

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,986 @@
...
+from nectarchain.makers.extractor.utils import CtapipeExtractor
+from nectarchain.utils import ContainerUtils
+
+# ---------------------------------------------------------------------------
+# Load bad pixel defaults
+# ---------------------------------------------------------------------------
+
+_BAD_PIXEL_FILE = (
+    Path(__file__).resolve().parent / "resources" / "bad_pixel_modules.json"
+)
```
pcorcam, line 64: There is a typo in the file name

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,986 @@
...
+    parser.add_argument(
+        "--bad_pixels", nargs="+", default=_DEFAULT_BAD_PIXELS, type=int
+    )
+    parser.add_argument("--use_bad_pixels", action="store_true", default=False)
+
+    parser.add_argument("--recompute_pedestal", action="store_true", default=False)
+    parser.add_argument("--recompute_gain", action="store_true", default=False)
+    parser.add_argument("--recompute_flatfield", action="store_true", default=False)
+    parser.add_argument("--recompute_charge", action="store_true", default=False)
+    parser.add_argument("--recompute_all", action="store_true", default=False)
```
pcorcam, line 141: Right now, these arguments don't force any recomputation. In fact, only `--recompute_gain` is passed via `_should_recompute()` through to `compute_gain` where the `overwrite` argument is adapted (L377). But even there, since there is a check to see whether `tool.output_path` exists (L402), this does not do anything.

So if we'd like to keep the option to force recomputations even if the output files exist already (which I think is a good idea), could you please update this?

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,986 @@
...
+                else FlatFieldSPEHHVStdNectarCAMCalibrationTool
+            )
+        else:
+            tool_class = (
+                FlatFieldSPENominalNectarCAMCalibrationTool
+                if self.args.free_pp_n
+                else FlatFieldSPENominalStdNectarCAMCalibrationTool
+            )
+
+        overwrite = self.args.gain_overwrite or self._should_recompute("gain")
```
pcorcam, line 379: Cf https://github.com/cta-observatory/nectarchain/pull/264/changes#r3511843567

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,986 @@
...
+
+        tool = tool_class(
+            progress_bar=True,
+            camera=self.args.camera,
+            run_number=run_number,
+            max_events=self.args.max_events_gain,
+            **kwargs,
+        )
+
+        if os.path.exists(tool.output_path):
```
pcorcam, line 404: Cf https://github.com/cta-observatory/nectarchain/pull/264/changes#r3511843567

src/nectarchain/trr_test_suite/observation_temperatures.py
```diff
@@ -0,0 +1,986 @@
...
+        existing = self._resolve_existing_output(
+            tool, "flatfield", run_number, self.flatfield_results
+        )
+        if existing:
+            return existing
+
+        self.log.info(f"Computing flatfield for run {run_number} (single pass)…")
+
+        tool.start()
+        tool.finish(return_output_component=True)
```
pcorcam, line 477: Missing `tool.setup()`

pcorcam (changes requested):
Hi @hashkar ,

Thanks a lot for the updates!

I found some small bugs and also some things that are still not really working as expected (see comments). Could you please have a look into those? Most importantly, for the next (and hopefully last) review request, please provide an example usage of the script that works as expected. In particular, it would be nice to obtain plots that don't show any "weird" behavior (cf [previous review](https://github.com/cta-observatory/nectarchain/pull/264#issuecomment-3953416267)).

As a side note, there now also exists a `HiloNectarCAMCalibrationTool` (see #303) which allows you to estimate the gain of the low-gain channel using flat-field events. I imagine you'd probably want to add this as well to correctly estimate the low gain of the pixels. But I'd suggest to make it part of a future PR once this one is merged.

Thanks again,
Pablo

PR #267, opened 2026-02-16
TRR : PMT transit time corrections

jlenain:
This PR voids the need to have the data file for the HV PMT transit time corrections at hand on one's own machine.


pcorcam (approved):
Thanks for this @jlenain, works great!

PR #269, opened 2026-02-18
Adjust min max colorbar for CameraDisplay plots

alessandromontanari:
Minor changes to adjust the colorbar range such that the displayed values are meaningful.

Now the values for bad pixels are always capped to 1, such that the camera display shows 0 or 1 for CAMERA-BADPIX 2D plots. 

Also, the position of the bad pixels is extracted as a boolean mask, and it is used to flag the pixels for the other 2D camera plots. The colorbar min-max range for these other plots is set to the min max values in the image w/o the bad pixels. The bad pixels are displayed in white, so that they do not interfere with the other meaningful values of the plotted quantity (see plot below for run 7214).

Screenshot from 2026-02-18 15-37-36

### DIFF
--- a/src/nectarchain/dqm/bokeh_app/tests/test_app_hooks.py
+++ b/src/nectarchain/dqm/bokeh_app/tests/test_app_hooks.py
@@ -61,12 +61,12 @@ def test_get_bad_pixels_position():
     mask_high, mask_low = get_bad_pixels_position(source, image_shape)
     assert mask_high.shape == image_shape
     assert mask_low.shape == image_shape
-    assert mask_high[1] == True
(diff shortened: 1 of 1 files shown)


jlenain:
> Hi @jlenain, the last commit should possible solve the exceptions, but I still have to add the data for the tests...

Hi @alessandromontanari ,

Many thanks ! I confirm the commit https://github.com/cta-observatory/nectarchain/pull/269/changes/957b48692014c5e90c6b9a1154b376aa9289bc21 beautifully solves the exception with the `CAMERA-BADPIX-PED-PHY-OVEREVENTS-HIGH-GAIN` dictionary key.

PR #285, opened 2026-03-06
ENF added to calculation of  NSB rate, new code for Laser source calibration added

medha1996c:


src/nectarchain/trr_test_suite/laser_source_intensity.py
```diff
@@ -0,0 +1,308 @@
...
+        charge[:, 0],
+        color="green",
+        yerr=np.transpose(std_err)[0],
+        marker="o",
+        linestyle="",
+        label="HG",
+    )
+    ax.errorbar(
+        intensity,
+        charge[:, 1] * ratio_hglg[5],
```
jlenain: Why are we taking the Hi/Lo ratio of the 6th run only here?

jlenain (changes requested):
Hi @medha1996c ,

Thanks a lot for this PR !
I have left a couple of comments to address.

src/nectarchain/trr_verification_package/utils.py
```diff
@@ -35,6 +35,8 @@
 
 # (filter, optical density, transmission)
 # configurations with the same order as in the document
+ENF = 1.12
```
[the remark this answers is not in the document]
medha1996c: Shifted to nectarchain.utils.constants and reference added

src/nectarchain/trr_verification_package/laser_source_intensity.py
```diff
@@ -0,0 +1,308 @@
...
+        charge[:, 0],
+        color="green",
+        yerr=np.transpose(std_err)[0],
+        marker="o",
+        linestyle="",
+        label="HG",
+    )
+    ax.errorbar(
+        intensity,
+        charge[:, 1] * ratio_hglg[5],
```
medha1996c: Any run can be taken in the linear region, updated with value from linear region. Hard-coding removed

medha1996c:
Rebased and modified codes pushed.

src/nectarchain/trr_verification_package/utils.py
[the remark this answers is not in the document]
jlenain: I meant add a reference to a paper or internal note as a comment into the code, such that in the long run, we remember where this value of 1.12 is coming from.

jlenain:
CI/docs failing, will be fixed by #376.

src/nectarchain/trr_verification_package/laser_source_intensity.py
```diff
@@ -0,0 +1,323 @@
...
```
jlenain, line 299: Again, why are we taking the Hi/Lo ratio of the 6th run only here? Hardcoding such an index is fragile, in the sense that if at some point the input run list has less than 6 runs, this code will break.

jlenain: Would it be sensible to use the average of the computed Hi/Lo ratio for all input runs instead ? Such as:

```python
        charge[:, 1] * np.mean(ratio_hglg),
```

PR #292, opened 2026-03-16
new test thermal code long run draft

ameclement:
Here is my draft: It is very long (runs for a long time also), and will not include every tool that is in here, as I am currently writing a Tool to collect everything I need in the Charges Container for the long runs. Although, I have found (and temporarily managed) some compatibilily issues with the run slicing. Everything is (hopefully) written down about how I proceeded. I will also send some additionnal files (for the monitoring database temperatures), and the tool, which I'll include in my final script. This will shorten the code and the runtime. Nonetheless, I still might have to run the pedestal tool outside of my tool, which will loop the events twice.


jlenain:
Hi @ameclement !

What is the status of this PR ? Is it work in progress ?

ameclement:
Hi @jlenain ! I apologize for the lack of response on my part, but it is still a work in progress. Everything works so far, but I didn't adapt it for the gui. I will do that ASAP and update you this week! Sorry again!

PR #293, opened 2026-03-20
Add tools to read the DB infos

vmarandon:
Add a set of tools to access SQLite db that contain monitoring information and transform it to a structure similar to the data structure from raw data files.


jlenain (approved):
Many thanks, @vmarandon !

PR #300, opened 2026-04-08
ToM Calculation

Cormed-Mack:
Changed the "window width" and "window shift" parameters of the tools that work on ToM calculation. This does not affect event rejection, or charge calculation (which is not needed for these tools). The result gives ToMs that have the same mean as those of the CubicSpline interpolation but with only slightly higher dispersion (+/- 0.2ns on high flasher intensities) I have included a supporting plot.
ToM Calculation Difference 7152 16V Flasher ctapipeonlycubic

ToM Calculation Difference 7144 8V Flasher ctapipeonlycubic


jlenain (approved):
Thanks a lot for this study, @Cormed-Mack !

PR #303, opened 2026-04-27
Add HiLo correction

pcorcam:
Added a `HiLoNectarCAMCalibrationTool` that computes HiLo corrections. See also `user_scripts/pcorrea/example_hilo.py` for a simple example. 

A required trait of this tool is a `gain_file`, which should contain a `GainContainer`. This `GainContainer` is copied, and its `low_gain` field is filled, which is the main purpose of the `HiLoNectarCAMCalibrationTool`. The copied container is written in a new `output_path` which is by default `/path/to/input/gain_file_hilo_corrected.h5` - so in the same directory as the input gain file.

The real work is performed in the `HiLoComponent`, and in particular in its `_compute_low_gain` method. Charges are computed exclusively for flat-field events, since we use the linear regime (`utils.constants.GAIN_LINEAR_RANGE = [7,400]` in photoelectrons, see [FEB characterization paper](https://arxiv.org/abs/2311.11631)) to compute the HiLo ratio. Note that for this to work properly, the pedestal needs to be subtracted when computing the charge. This should be done by passing a `pedestal_file` to the `ChargesComponent` subcomponent of the tool. 

Note that the `gain_container` that is copied and written to the `output_path` is completed with values for the entire camera, even if the pixels are missing from the run (this is achieved with `ContainerUtils.add_missing_pixels_to_container`).  For missing pixels, or pixels were the HiLo computation fails, the default `utils.constants.HILO_DEFAULT` is used. 

A complete unit-test file is added for the `HiLoComponent` (`test_hilo_component.py`); for the `HiLoNectarCAMCalibrationTool` (`test_hilo_makers.py`) currently only the `output_path` is tested. The loading of charge containers is exactly the same as done in other `GainNectarCAMCalibrationTool`s. 

An example of the HiLo coefficients computed for some (miscellaneous) runs is shown below. Note that the HiLo coefficients are not stored directly; you can get them by loading the `gain_container` from the `output_path` and then just divide `gain_container["high_gain"][:,0] / gain_container["low_gain"][:,0]`.

image


jlenain (commented):
Hi @pcorcam ,

Many many thanks for this PR !

Would it be possible please to add in the example user script the prerequisite parts to process the pedestal computation for run 6249 as well as the SPE fit computation for run 3936 ?

Thanks !

src/nectarchain/user_scripts/pcorrea/example_hilo.py
```diff
@@ -0,0 +1,62 @@
...
+
+# Set NECTARCAMDATA environment
+os.environ["NECTARCAMDATA"] = "/data/users/pcorrea"
+
+# Input gain file and pedestal file
+dir_calib_files = "./calibration_files/"
+pedestal_file = dir_calib_files + "pedestal_6249.h5"
+gain_file = (
+    dir_calib_files + "FlatFieldSPENominalStdNectarCAM_run3936_maxevents50000"
+    "_LocalPeakWindowSum_window_shift_4_window_width_16.h5"
```
jlenain: One small detail here: the gain is assumed to be computed using the `LocalPeakWindow Sum` between [-4; +12] ns around the pulse peak, while below (line 25, https://github.com/cta-observatory/nectarchain/pull/303/changes#diff-c7e5ca5bc5d3fe40e436199f6d299a177a7e176cbf61c7da1d3ea871c3483c92R25), the time window [-4; +4] ns is used for the Hi/Lo computation.

pcorcam: Thanks for catching that! I've updated the example script such that the charge extraction with the HiLo tool is always the same as with the SPEfit tool.

pcorcam:
Hi @jlenain ! 

I've updated the `example_hilo.py` so that it can now be run stand-alone. So if the pedestal and/or gain files are not present, the corresponding tools will be run.

Let me know if this looks good to you!

jlenain (approved):
Thanks a lot, @pcorcam !

PR #304, opened 2026-04-28
Handle outliers in z-range for camera plots

alessandromontanari:
refactor(src/nectarchain/dqm/bokeh_app):
- app_hooks.py: adapt z-range of camera displays to handle outliers

Small refactor in app_hooks to keep 99.5% of the events inside the z-range, so as to avoid outliers shadowing the bulk of the data.


jlenain (approved):

PR #307, opened 2026-04-28
Pixels vals histograms and vs ids shown for all camera displays 

alessandromontanari:
feat(src/nectarchain/dqm/bokeh_app):
- app_hooks.py: implement make_pixels_vals_histo() with stats and make_pixel_val_vs_id() added to make_camera_displays(); update update_camera_displays() and update_timelines() for the frontend output; remove reset of camera displays and timelines when updating to new run -- obsolete
- main.py: adapt layouts to accommodate more plots in rows


alessandromontanari:
The tests still need to be updated with these changes: I'm on it...

jlenain (approved):
Just a curiosity we just discussed and I can't understand, I just stumbled across a particular case (run 420 with the QM camera) for which some camera displays are empty, but the histogram and scatter plot show some data. Just some meat for thought for later.

PR #277, opened 2026-03-02
Temperature dependent code modifications

medha1996c:
This pull request is to develop a centralized script for temperature dependent codes : 

1. Linearity : produce plots of a)fit parameters 'a' and 'b' and  b)HG/LG as function of temperature for different  transmissions.
2. Charge resolution: Charge resolution at a given temperature as a function of temperature, fit with temperature and   produce %change/ deg C value. Check for different nsb levels
3. Deadtime : a) Deadtime as a function of temperature for different intensities b) Relative difference between Collected trigger rate and measured trigger rate as function of temperature (TRR requirement : to be less than 10 % ).

Input run file : resources/observation_temperature_run_list.json

PR #311, opened 2026-05-04
Add test for deadtime and event time-difference distributions to Acceptance Verification

alessandromontanari:
feat(src/nectarchain/acceptance_verification_package):
- gui.py: refactor to add deadtime test
- deadtime.py: implement placeholder for deadtime test following TRR test suite logic

This PR will implement the test to plot deadtime and event time-difference distributions for the acceptance verification on site. It follows the logic already implemented for the TRR test suite.


src/nectarchain/acceptance_verification_package/deadtime.py
```diff
@@ -0,0 +1,270 @@
...
+        plot_fitted_rate_vs_collected_trigger_rate(
+            np.array(ids),
+            np.array(collected_trigger_rates),
+            np.array(fitted_trigger_rates),
+            np.array(fitted_trigger_rates_err),
+            sources_label,
+            output_dir,
+            temp_output,
+        )
+
```
tibaldo: Perhaps add also a band or line that visualizes the requirement B-TEL-1270 (+B-MST-1280), i.e. deadtime < 5% for trigger rates up to 7 MHz.

alessandromontanari: Yes, will do. This is still a draft PR anyway because I haven't fully tested it.

alessandromontanari:
Hi @tibaldo, with these last few pushes, this PR satisfies the same requirements as for the TRR deadtime test, and produces the plots for the GUI in `acceptance_verification_package`. With the CLI: 
```
python deadtime.py --source 0 0 0 0 0 0 0 0 0 0 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 32 -r 3332 3333 3334 3335 3336 3337 3338 3339 3340 3341 3342 3343 3344 3345 3346 3347 3347 3348 3349 3350 3552 3553 3554 3555 3556 3557 3558 3559 3560 3561 3562 -c NectarCAMQM --evts 4000 --output ./
```
you can get the plots below: 

deadtime_and_expo_fit_3556

deadtime_pc_vs_trigger_rate

fitted_vs_collected_trigger_rates

You can also visualise the same plots via the GUI (in which you'll see the distributions for all the runs).

src/nectarchain/trr_test_suite/resources/source_type_labels.json
```diff
@@ -0,0 +1,13 @@
+{
+    "0": {"source": "Flatfield", "color": "red"},
+    "1": {"source": "Single Photo Electron", "color": "blue"},
```
tibaldo, line 3: This is possibly a naive question: what is the source of this numbering convention? Is there a document you could quote in the comments to let people know where to look for information? Why is this specific to the acceptance verification package and not general to nectarchain/ctapipe_io_nectarcam?

src/nectarchain/acceptance_verification_package/deadtime.py
```diff
@@ -0,0 +1,308 @@
...
+            32,
+            32,
+            32,
+            32,
+            32,
+        ],
+    )
+    parser.add_argument(
+        "-e",
+        "--evts",
```
tibaldo: Call `max_events` for transparency and consistency with other nectarchain tools?

src/nectarchain/acceptance_verification_package/deadtime.py
```diff
@@ -0,0 +1,308 @@
...
+        "--log",
+        help="log level",
+        default="info",
+        type=str,
+    )
+
+    return parser
+
+
+def main():
```
tibaldo: This function looks very similar to main in `trr_test_suite.deadtime.py`. Why do you need to re-implement a tool in the `acceptance_verification_package` rather than inheriting from  the one in `trr_test_suite.deadtime.py` so that you avoid code duplication?

src/nectarchain/acceptance_verification_package/gui.py
```diff
@@ -2,14 +2,15 @@
 
 from PyQt5.QtWidgets import QApplication
 
-from nectarchain.acceptance_verification_package import hillas_validation
+from nectarchain.acceptance_verification_package import deadtime, hillas_validation
```
tibaldo: Could you import here directly the tool in `trr_test_suite` rather than creating a new one?

tibaldo (commented):
Ciao @alessandromontanari, thank you very much for proposing this PR! I look through it and I have some questions mostly related to keep consistency across the NectarCAM software and avoid code duplication. Otherwise this looks good to merge to me!

src/nectarchain/trr_test_suite/resources/source_type_labels.json
alessandromontanari: Hi @tibaldo, the numbers come from `ctapipe.containers.EventType`. I just pushed a commit where I added a docstring to the `get_labels()` function in `deadtime.py` explaining where the ids and labels come from. 
You're right, though, this should probably not be specific to the acceptance_verification_package only, but more general... Maybe, the function `get_labels()` could be moved to `src/nectarchain/utils/constants.py`. But then, the JSON file too should be moved outside of acceptance_verification_package. I guess we could discuss this at the next software meeting?

src/nectarchain/acceptance_verification_package/deadtime.py
alessandromontanari: In principle, I would agree on this. However, `--evts` is used by several tools in the trr_test_suite (deadtime, linearity, pedestal, etc.). That is basically where this comes from...

src/nectarchain/acceptance_verification_package/deadtime.py
alessandromontanari: Technically, you are right. It is very similar. I still imported a few tools defined in `trr_test_suite.deadtime.py` into `acceptance_verification_package.deadtime.py` to avoid duplicating them at least. For backward compatibility, I would still prefer to redefine the main function in `acceptance_verification_package.deadtime.py`, because `trr_test_suite.deadtime.py` is also used in #277. I prefer to avoid some nightmare scenarios with conflict with that PR... Nevertheless, I agree that this can be optimised, and I believe it could be done by changing only a few lines of code. Shall we open an issue on this so that I will be reminded to work on it in the future?

src/nectarchain/acceptance_verification_package/gui.py
alessandromontanari: The answer for now is no, for what I commented above. But it will likely become feasible once we merge #277, and I won't need to worry about potential merge conflicts anymore.

src/nectarchain/trr_test_suite/resources/source_type_labels.json
tibaldo: Okay, let's leave it like that then and we can have another PR after discussion to fix this across nectarchain.

src/nectarchain/acceptance_verification_package/deadtime.py
tibaldo: Ok, another thing to fix in the future to make it consistent across nectarchain then

src/nectarchain/acceptance_verification_package/deadtime.py
tibaldo: Maybe @jlenain can chime in to say if in his opinion it's better to go ahead with the implementation in `acceptance_verification_package` as you propose to avoid conflicts with other PR or we can try to unify this already?

jlenain: Hi @alessandromontanari , @tibaldo !

I think the question is: are the files `trr_test_suite/deadtime.py` and `acceptance_verification_package/deadtime.py` different ? As I understand it, they should be the same, so my point is that `acceptance_verification_package/deadtime.py` should not exist. The one in `trr_test_suite` would be enough, and I will definitely help in case of conflicts when merging PR #277 (I am used to it, and Pycharm helps me a lot with this).

alessandromontanari: Hi @jlenain! 
Ok, then, I shall rework this PR to try and import `deadtime.py` directly from `trr_test_suite`. @tibaldo, I'll try to update the PR this week.

tibaldo: Thank you very much @alessandromontanari !

alessandromontanari:
Hi @tibaldo and @jlenain, 
I removed `deadtime.py` in `acceptance_verification_package`, and reworked its cousin in `trr_test_suite`. Now, the `gui.py` in `acceptance_verification_package` imports `deadtime.py` from `trr_test_suite`. When you run that GUI, you should get the same results as shown above if you set the test_type argument to 'av'. These latest changes may create conflicts with #277. Nevertheless, I reckon it should not be anything too troublesome. I also moved `source_type_labels.json`  from `acceptance_verification_package/resources` to `trr_test_suite/resources` as a temporary solution.

tibaldo:
Thank you @alessandromontanari, after the latest changes everything looks good as far as I can tell!

tibaldo (approved):

PR #312, opened 2026-05-04
Fix plot colorbar labels in Bokeh app

alessandromontanari:
refactor(src/nectarchain/dqm/bokeh_app):
- data/labels.json: all camera display color bar labels set to ADC counts

This PR sets the colorbar labels for the camera display plots always to ADC counts, as there is no conversion to p.e. in the processors.


jlenain (commented):
Hi @alessandromontanari ,

Thanks a lot, you're right, the DQM only uses R0 data at the moment. Do you think it would still make sense to make the mention "Pedestal" and "Average signal" in the corresponding labels ?

jlenain (commented):
Hi @alessandromontanari ,

Thanks a lot, you're right, the DQM only uses R0 data at the moment. Do you think it would still make sense to make the mention "Pedestal" and "Average signal" in the corresponding labels ?

alessandromontanari:
> Hi @alessandromontanari ,
> 
> Thanks a lot, you're right, the DQM only uses R0 data at the moment. Do you think it would still make sense to make the mention "Pedestal" and "Average signal" in the corresponding labels ?

Hi @jlenain, yes I thought about that. But what if we keep the colorbar labels for only the units, and instead, rework the plot titles? Because "pedestal" and "average signal", for instance, are already in the plot titles somehow -- even though maybe not really clear for the user yet. I could add another entry in the labels.json for more explanatory titles, corresponding to each key in the dictionary.

jlenain:
> > Hi @alessandromontanari ,
> > Thanks a lot, you're right, the DQM only uses R0 data at the moment. Do you think it would still make sense to make the mention "Pedestal" and "Average signal" in the corresponding labels ?
> 
> Hi @jlenain, yes I thought about that. But what if we keep the colorbar labels for only the units, and instead, rework the plot titles? Because "pedestal" and "average signal", for instance, are already in the plot titles somehow -- even though maybe not really clear for the user yet. I could add another entry in the labels.json for more explanatory titles, corresponding to each key in the dictionary.

Hi @alessandromontanari , yes, good idea for the last proposal, it would be clearer for users, I agree.

jlenain:
Hi @alessandromontanari !

Sorry to come back so late to this PR. I understood it would still need some changes (see the discussion above about labels and plot titles). Is it still the case or could we merge it ?

Thanks !

alessandromontanari:
> Hi @alessandromontanari !
> 
> Sorry to come back so late to this PR. I understood it would still need some changes (see the discussion above about labels and plot titles). Is it still the case or could we merge it ?
> 
> Thanks !

Hi @jlenain, I think we can merge it. I wanted to come back on this once the database is reprocessed with the new DQM features. So that I implement labels and fix the titles once I see all the plots available in the database.

jlenain (approved):
Thanks @alessandromontanari !

PR #315, opened 2026-05-12
Add waveform plots and tab to Bokeh

alessandromontanari:
This PR adds waveform plots for all the gains and for both physical and pedestal events to the Bokeh web app. At the moment, only mean waveforms are displayed. In the future, waveforms for all the channels will also be displayed.


jlenain (approved):
Many thanks, @alessandromontanari !

Issue #314, opened 2026-05-05
Use enumerator comparison for event types in the DQM


alessandromontanari:
**Please describe the use case that requires this feature.**
In all processors in the DQM, the event types from `evt.trigger.event_type.value` are still compared to hard-coded numbers. 

**Describe the solution you'd like**
We should use an enumerator comparison to `EventType` in `ctapipe.containers` in all processors.

**Describe alternatives you've considered**
No other solution should be applied, as this one also respects the logic of how ctapipe is written, and nectarchain is based on ctapipe.

**Additional context**

PR #316, opened 2026-05-22
Save waveforms for all pixels extracted by the DQM in the database

alessandromontanari:
feat(src/nectarchain/dqm):
- __init__.py: change processor init name for waveforms
- waveforms.py: moved from mean_waveforms.py, include all waveforms (not only one averaged over all pixels) in the dictionary saved to database
- start_dqm.py: minor fixes following changes in waveforms processor script
- tests/test_waveforms.py: fixed tests

This PR includes the saving of the waveforms for all the pixels to the database, not only the one averaged over all pixels. For consistency, the processor has been renamed `waveforms.py`, from `mean_waveforms.py`. `__init__.py`, `start_dqm.py`, and `test_waveforms.py` have been updated accordingly.

### DIFF
--- a/src/nectarchain/dqm/waveforms.py
+++ b/src/nectarchain/dqm/waveforms.py
@@ -1,0 +1,172 @@
+import os
+
+import numpy as np
+from matplotlib import pyplot as plt
+
+from .dqm_summary_processor import DQMSummary
+
+__all__ = ["WaveFormsHighLowGain"]
+
+
+class WaveFormsHighLowGain(DQMSummary):
+    def __init__(self, gaink, r0=False):
+        self.k = gaink
+        self.Pix = None
+        self.Samp = None
+        self.tel_id = None
+        self.wf = None
+        self.wf_ped = None
+        self.counter_evt = None
+        self.counter_ped = None
+        self.wf_average = None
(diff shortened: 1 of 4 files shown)


src/nectarchain/dqm/waveforms.py
```diff
@@ -0,0 +1,172 @@
...
+        self.counter_evt = 0
+        self.counter_ped = 0
+
+        self.wf_average = np.zeros((Pix, Samp))
+        self.wf_ped_average = np.zeros((Pix, Samp))
+
+        self.wf_list_plot = list(np.arange(1, Samp + 1))
+
+    def process_event(self, evt, noped):
+        if evt.trigger.event_type.value == 32:  # only peds now
```
jlenain, line 67: Should we try to use an `EventType` enum here, instead of hardcoding a value ?

alessandromontanari: Hi @jlenain, yes, I spotted that too! These comparisons appear several times in the DQM, with hardcoded values in several processors. I was actually thinking of solving them all in a separate PR, which is also the reason why I created #314 a while ago.

jlenain (approved):

PR #317, opened 2026-05-28
Fix `FlatFieldComponent` so that it can be used for runs with an incomplete camera

pcorcam:
This is a _temporary_ ugly workaround so that the `FlatFieldComponent` can correctly treat runs without a complete camera.

For any `ArrayDataComponent` (such as a `WaveformsComponent` or `ChargesComponent`), waveforms are naturally only selected for the available pixels in the camera (see [here](https://github.com/cta-observatory/nectarchain/blob/2c5f079eb1c8742ccf0f7a64e7963269bdfac91f/src/nectarchain/makers/component/core.py#L55)). This PR tries to mimic that during the waveform extraction and manipulation in the `FlatFieldComponent`, basically by just selecting the `self.pixels_id` which are defined by the parent `NectarCAMComponent` (where it is already taken into account which pixels were used for the run). 

To test that it's indeed working, use the example script of @ArmelleJB (`user_scripts/ajardinb/flatfield_example.py`) as follows, for [run 6253](http://nectarcam.in2p3.fr/elog/nectarcam-data-qm/?mode=full&reverse=0&reverse=1&npp=20&Subject=6239) which contains 1544 pixels:

`python flatfield_example.py -r 6253 -m 1000`
**NOTE**: You should set `window_shift = 4` in the script, otherwise all pixels will be tagged as bad in the first run. I also checked that this works by specifying a `pedestal_file` and `gain_file` in the script. 

In any case, this whole manipulation of waveforms is only necessary so that we can apply the two-pass method in `flatfield_example.py`. However, at some point we can just start using the existing gain tools to compute the gain, making the two-pass method obsolete. When we decide to do that, we should just use a `ChargesComponent` subcomponent in `FlatFieldComponent.__call__()` (like done in most other components), and perform the `eff_coef` computations in `FlatFieldComponent.finish()`. That way we also won't have to save all the charges in the output of the `FlatFieldComponent`. Just something to keep in mind :)

### DIFF
--- a/src/nectarchain/makers/component/flatfield_component.py
+++ b/src/nectarchain/makers/component/flatfield_component.py
@@ -229,7 +229,7 @@ of the waveform"
             )
 
             # get the waveform
-            wfs = event.r0.tel[self.tel_id].waveform
+            wfs = event.r0.tel[self.tel_id].waveform[:, self.pixels_id]
 
             # subtract pedestal container if filled
             # otherwise use the mean of the n first samples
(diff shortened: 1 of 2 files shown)


jlenain:
Hi @ArmelleJB !

If you feel comfortable with it, may I ask you please to review this pull request ?

Many thanks in advance !

ArmelleJB (approved):

PR #318, opened 2026-06-01
Add detailed HoverTool to Bokeh app

alessandromontanari:
This PR adds detailed HoverTools both for camera display plots and the corresponding scatter plots. The tools are defined separately in dedicated functions. Unit tests for the newly defined functions are also created, and they pass in local.


jlenain (approved):
So cool !
Thank you very much, @alessandromontanari !

PR #320, opened 2026-06-02
Fix START-TIMES not writing to fits in the DQM

alessandromontanari:
fix(src/nectarchain/dqm):
- dqm_summary_processor.py: refactor _create_hdu to fix bug not writing dict objects to the database

This PR fixes `_create_hdu` in `dqm_summary_processor.py`, to write all the DQM output to the fits files (and thus to the database later).

### DIFF
--- a/src/nectarchain/dqm/dqm_summary_processor.py
+++ b/src/nectarchain/dqm/dqm_summary_processor.py
@@ -51,38 +51,41 @@ class DQMSummary:
 
     @staticmethod
     def _create_hdu(name, content):
-        # Convert to numpy array
-        arr = np.asarray(content)
+        data = Table()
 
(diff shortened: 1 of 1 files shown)


alessandromontanari:
Hi @jlenain this should fix what we discussed in private.
With 
`python start_dqm.py -c NectarCAMQM --noped -r 7213 --max-events 500 $NECTARCAMDATA $NECTARCAMDATA`
I get this:
Screenshot from 2026-06-02 16-03-29
Even though the format of CAMERA-TEMPERATURE-TREND is different (it is because we need an ImageHDU for a 2D array), I checked separately with another script, and the content of the fits file for that entry is correct.

jlenain (approved):
Thanks a lot for that, @alessandromontanari !

PR #322, opened 2026-06-08
Comparisons for event_type values in the DQM to EventType

### LINKED ISSUE #314: Use enumerator comparison for event types in the DQM

alessandromontanari:
This PR addresses issue #314. All comparisons for `evt.trigger.event_type.value` in processors of the DQM are now done with the enum `EventType` from `ctapipe.containers`, instead of hardcoded values. 
However, the comparison is only implemented (for now) for the SKY_PEDESTAL and SUBARRAY event types because the DQM does not yet generate plots for the other types. This will be addressed later.


src/nectarchain/dqm/tests/test_mean_waveforms.py
jlenain, line 1: Hi @alessandromontanari !

Why has this test file been deleted in this PR ? Was it on purpose, or an oversight ? Thanks !

src/nectarchain/dqm/mean_waveforms.py
```diff
@@ -64,7 +65,7 @@ def configure_for_run(self, path, Pix, Samp, Reader1, **kwargs):
         self.wf_list_plot = list(np.arange(1, Samp + 1))
 
     def process_event(self, evt, noped):
-        if evt.trigger.event_type.value == 32:  # only peds now
+        if evt.trigger.event_type == EventType.SUBARRAY:  # only peds now
```
jlenain: Oups, I think I did something wrong here, it should be the opposite logic...

src/nectarchain/dqm/mean_waveforms.py
```diff
@@ -77,7 +78,7 @@ def process_event(self, evt, noped):
                 # (1855,60), or 3D (2, 1855, 60) for 2-gain channels or
                 # (1, 1855, 60) for single-gain channel
                 waveform = evt.r1.tel[self.tel_id].waveform[..., ipix, :]
-            if evt.trigger.event_type.value == 32:  # only peds now
+            if evt.trigger.event_type == EventType.SUBARRAY:  # only peds now
```
jlenain: Same here: Oups, I think I did something wrong here, it should be the opposite logic...

src/nectarchain/dqm/mean_waveforms.py
jlenain: Actually, I think the logic was flawed here from the start, it should be the other way around, isn't it ?

```python
if evt.trigger.event_type != EventType.SUBARRAY:  # only peds now
    self.counter_ped += 1
else:
    self.counter_evt += 1
```
or even better:

```python
if evt.trigger.event_type == EventType.SKY_PEDESTAL or evt.trigger.event_type == EventType.DARK_PEDESTAL or evt.trigger.event_type == EventType.ELECTRONIC_PEDESTAL:  # only peds now
    self.counter_ped += 1
else:
    self.counter_evt += 1
```

src/nectarchain/dqm/tests/test_mean_waveforms.py
alessandromontanari: Hi @jlenain, no the test for the waveforms is now in `test_waveforms.py`, which was added when #316 was merged. I basically changed the script names for waveforms from `_mean_waveforms` to `_waveforms` I think I never pushed this deleted file, so I did it here. Although now I'm not too sure it is needed.

src/nectarchain/dqm/mean_waveforms.py
alessandromontanari: Hi @jlenain, we were using hardcoded 32 for pedestals before. But 32 in `ctapipe.containers.EventType` is for standard physics stereo trigger right?
And to answer your second comment, yes, your second solution is likely the best. We can directly implement it. But in that way, we will have the same labels for all the pedestal plots, i.e. SKY_PEDESTAL, DARK_PEDESTAL and ELECTRONIC_PEDESTAL. Are we ok with that? Otherwise, we need to implement different counters for each pedestal, and later maybe also for other event types appearing in `ctapipe.containers.EventType`.

jlenain: Hi @alessandromontanari ,
I would tend to say, let's go that way for the moment. At the very least, this would fix the mix between (any kind of) pedestal and physical events.

jlenain (approved):

PR #326, opened 2026-06-08
Completion of the category-A calibration pipeline

pcorcam:
This PR combines the work of #207 (writing category-A calibration files) and #226 (running calibration tools in sequence). Essentially, the `CalibrationWriterNectarCAM` tool is now "absorbed" into the `finish()` method of the `PipelineNectarCAMCalibrationTool`. 

The main workflow is as follows (see also the example script `user_scripts/pcorrea/example_pipeline.py`):

- Perform a detailed setup of the different calibration tools.
- Run the calibration tools in sequence. The output of the pedestal and HiLo-corrected gain tools are used as input for the flat-field tool.
- Read the `NectarCAMContainer`s from the output files of each subtool and "translate" them to the relevant `ctapipe` containers. 
- Write the category-A calibration file in `.fits`, `.fits.gz`, or `.h5` format (this part is copied from `lstcam_calib`). 

Notable changes/additions are the following:

- HiLo corrections are included in the calibration pipeline.
- Any pixel tagged as bad in any form (e.g. missing in hardware or tagged as bad during a calibration step) is tagged as an `unusable_pixel` in the `WaveformCalibrationContainer`. Any `unusable_pixel` gets given default values for all calibration coefficients (pedestal, gain, flat-field) in the final output.
- The `save_tmp` trait gives the choice whether to keep the output files of all subtools.
- The `all_default` trait allows to quickly generate a cat-A file filled with default values.

Some things to note for the future:

- We'll need to adapt our `NectarCAMContainer`s to the "standard" `ctapipe` containers. This will avoid doing the ugly manual stuff that we have to do now to "translate" the containers in order to write the calibration files. But this will probably make sense when we merge to a `ctapipe` dependence beyond `v0.24`. 
- We need to find a way to properly tag bad pixels and keep track of them when passing through the entire calibration pipeline. On top of that, the way each `NectarCAMContainer` tags a bad pixel is different, so this should also be uniformed. 
- The `FlatFieldContainer` of `nectarchain` saves charges in its output file. This is obsolete by this time, so the `FlatFieldNectarCAMCalibrationTool` should be revisited.


src/nectarchain/makers/calibration/calibration_pipeline.py
```diff
@@ -67,148 +134,758 @@ class PipelineNectarCAMCalibrationTool(NectarCAMCalibrationTool):
...
+        self._nectarcam_containers["flatfield"].event_id = np.ones((1,))
+
+        # FF charges (dummy value)
+        self._nectarcam_containers["flatfield"].amp_int_per_pix_per_event = np.ones(
+            (1, N_GAINS, N_PIXELS)
         )
 
-    def _setup_config(self):
+        # Default timestamps (my birthday :)
+        default_timestamp = 791054100000000000  # ns
```
jlenain, line 416: 😁

src/nectarchain/makers/calibration/calibration_pipeline.py
```diff
@@ -67,148 +134,758 @@ class PipelineNectarCAMCalibrationTool(NectarCAMCalibrationTool):
...
+            # These will be tagged by either NaN or default values
+            mask_ped = np.logical_or(
+                np.isnan(pedestal),
+                pedestal == PEDESTAL_DEFAULT,
+            )
+            hardware_failing_pixels[mask_ped] = True
+
+            # Check for hardware failing pixels in gain container
+            # Only need to check high gain
+            gain = self._nectarcam_containers["gain"].high_gain[..., 0]
```
jlenain, line 555: 0 -> constant `HIGH_GAIN` ?

src/nectarchain/makers/calibration/calibration_pipeline.py
```diff
@@ -67,148 +134,758 @@ class PipelineNectarCAMCalibrationTool(NectarCAMCalibrationTool):
...
+
+        FF_pixel_mask = eff_coef_per_pixel_per_event == 0
+
+        # Mask bad pixels for FF coefficient computations
+        FF_coef_per_pixel_per_event_masked = np.ma.masked_array(
+            1.0 / eff_coef_per_pixel_per_event,
+            mask=np.broadcast_to(FF_pixel_mask, eff_coef_per_pixel_per_event.shape),
+        )
+
+        # Compute mean, median, std of FF coefficients, for bad pixels fill with 0
```
jlenain: Shouldn't these values be filled with `np.nan` instead?

src/nectarchain/makers/calibration/calibration_pipeline.py
```diff
@@ -67,148 +134,758 @@ class PipelineNectarCAMCalibrationTool(NectarCAMCalibrationTool):
...
+
+                # Workaround for astropy#17930, attach missing units
+                for col, value in self._ctapipe_containers[key].items():
+                    if unit := getattr(value, "unit", None):
+                        t[col].unit = unit
+
+                hdul.append(fits.BinTableHDU(t, name=key))
+
+            hdul.writeto(self.output_path, overwrite=self.overwrite)
+
```
jlenain, line 887: Add an `else` statement here, gracefully raising an exception if the output format is neither HDF5 nor FITS ?

src/nectarchain/makers/calibration/calibration_pipeline.py
pcorcam: This zero is not related to a gain axis; there is a separate `GainContainer.low_gain` attribute. The zero is just from the fact that the `GainContainer.high_gain` attribute has a shape `(N_pixels,3)`. Of the second axis, the `0`th element is the best-fit gain value, while the `1`st and `2`nd elements correspond to its errors.

src/nectarchain/makers/calibration/calibration_pipeline.py
pcorcam: If you have a strong opinion, I can add it (I just copied this part from the original `lstcam_calib` code). In any case, the `output_format` trait is of type `CaselessStrEnum`, which will yield an error if not one of the formats defined in `OUTPUT_FORMATS = = [".h5", ".fits", ".fits.gz"]`. So there will never be a point where `self._write_catA_calibration_file()` will be called with a wrong output format.

src/nectarchain/makers/calibration/calibration_pipeline.py
pcorcam: Yes good catch - this is a relic from the old version! I will update it right away.

src/nectarchain/makers/calibration/calibration_pipeline.py
jlenain: Absolutely, my bad !

src/nectarchain/makers/calibration/calibration_pipeline.py
jlenain: Ok, indeed, that would be good enough. Thanks!

jlenain:
Thanks a lot, @pcorcam !

Just a note for the future: the charge computation part always gives warnings, because HDF5 charge files are actually not stored on disk in the process.

jlenain (approved):

PR #331, opened 2026-06-12
Ping pong monitoring for dqm

ameclement:
this is the processor used for the trigger missing modules, the ping-pong monitoring. it is implemented with all the changes in start_dqm.py and __init__.py. It should plot a camera with 0 changes or above 0 if modules skip a trigger and are in a different state than the general event. it should save the times where it happens in the fits file.

### DIFF
--- a/src/nectarchain/dqm/ping_pong.py
+++ b/src/nectarchain/dqm/ping_pong.py
@@ -49,7 +49,7 @@ class PingPongMonitoring(DQMSummary):
             EngineeringCameraFrame()
         )
         self.cmap = "gnuplot2"
-        self.pixel_ids = Reader1.camera_config.pixel_id_map
+        self.pixel_ids = np.arange(self.Pix)
         self.subarray = Reader1.subarray
 
         self.change = np.zeros(len(self.pixel_ids))
--- a/src/nectarchain/dqm/ping_pong.py
+++ b/src/nectarchain/dqm/ping_pong.py
@@ -54,17 +54,17 @@ class PingPongMonitoring(DQMSummary):
 
         self.change = np.zeros(len(self.pixel_ids))
         for i, evt1 in enumerate(Reader1):
-            # log.warning("configure the run with the first event")
             self.run_start1 = evt1.nectarcam.tel[self.tel_id].svc.date
             cell_id = evt1.nectarcam.tel[self.tel_id].evt.first_cell_id
             event_id = evt1.index.event_id
             trigger_time = evt1.trigger.time.value
             ping = (cell_id & 0x400).astype(bool)
+            # This performs a bitwise AND operation between
+            # the value of cell_id and the hexadecimal number 0x400
+            # 0x400 in binary is 010000000000 (12 bits)
+            # checks if the 11th bit (from the right, starting at 0)
+            # is set in cell_id
             if event_id == 1 or i == 0:
-                # log.warning(
-                # f"collecting reference data from the first event,"\
-                # f"event {event_id}"
-                # )
                 self.ref_parity = evt1.index.event_id % 2
                 self.ref_state = np.array(ping)
                 self.first_event = np.array(ping.view(np.int8))
@@ -79,7 +79,6 @@ class PingPongMonitoring(DQMSummary):
                     self.change[mismatches] += 1
                     self.event_times.append(trigger_time)
                     self.nchanges += 1
-            # log.warning(f"the parity of the first event is {self.ref_parity}")
 
     def process_event(
         self,
@@ -96,6 +95,8 @@ class PingPongMonitoring(DQMSummary):
 
         self.event_id.append(trigger_id)
         if not np.array_equal(state, expected):
+            # TODO: add a comment explaining
+            # why only logging the first ten elements
             log.warning(
                 f"Mismatch: Event {trigger_id}, State: {state[:10]}"
                 f" (expected {expected[:10]}), time={trigger_time}"
@@ -106,6 +107,8 @@ class PingPongMonitoring(DQMSummary):
             self.change[mismatches] += 1
             self.event_times.append(trigger_time)
             self.nchanges += 1
+            # TODO: add a comment explaining
+            # why only logging the first ten elements
             log.warning(
                 f"Reset reference. Changes incremented at indices: {mismatches[:10]}..."
             )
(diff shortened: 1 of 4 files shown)


src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+        self.camera = Reader1.subarray.tel[self.tel_id].camera.geometry.transform_to(
+            EngineeringCameraFrame()
+        )
+        self.cmap = "gnuplot2"
+        self.pixel_ids = Reader1.camera_config.pixel_id_map
+        self.subarray = Reader1.subarray
+
+        self.change = np.zeros(len(self.pixel_ids))
+        for i, evt1 in enumerate(Reader1):
+            # log.warning("configure the run with the first event")
```
alessandromontanari: Salut @ameclement! Thanks a lot for working on this PR! I am running the DQM locally before approving the PR, but it seems fine so far. I only have a couple of very minor comments. Starting from this line. Do we need to keep this line with the `log.warning`? If it is not necessary, I'd suggest removing it. Indeed, we may not want to have either a `log.info` or a `log.warning` there, because this would be printed for all the events in the enumerate...

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+            self.run_start1 = evt1.nectarcam.tel[self.tel_id].svc.date
+            cell_id = evt1.nectarcam.tel[self.tel_id].evt.first_cell_id
+            event_id = evt1.index.event_id
+            trigger_time = evt1.trigger.time.value
+            ping = (cell_id & 0x400).astype(bool)
```
alessandromontanari, line 61: I admit this is due to my ignorance, but I had to look up what the hexadecimal number `0x400` corresponds to. I see the line is to convert a bitwise AND operation to bool, and you are checking if the 11th bit (from the right, starting at 0) is set in `cell_id`. Could you maybe write a comment above this line explaining this check?

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+            if event_id == 1 or i == 0:
+                # log.warning(
+                # f"collecting reference data from the first event,"\
+                # f"event {event_id}"
+                # )
```
alessandromontanari: Same comment for this commented `log.warning` as what I commented before

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+                if len(pop1) != 0 and len(pop1) != len(self.pixel_ids):
+                    mismatches = min([pop1, pop2], key=len)
+                    log.warning(
+                        f"The first event has some discreptancies"
+                        f" for pixels {mismatches}"
+                    )
+                    self.change[mismatches] += 1
+                    self.event_times.append(trigger_time)
+                    self.nchanges += 1
+            # log.warning(f"the parity of the first event is {self.ref_parity}")
```
alessandromontanari: Same comment for this commented `log.warning` as what I commented before

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+        ping = (cell_id & 0x400).astype(bool)
+        parity = trigger_id % 2
+        state = np.array(ping)
+        expected = self.ref_state if parity == self.ref_parity else ~self.ref_state
+
+        self.event_id.append(trigger_id)
+        if not np.array_equal(state, expected):
+            log.warning(
+                f"Mismatch: Event {trigger_id}, State: {state[:10]}"
+                f" (expected {expected[:10]}), time={trigger_time}"
```
alessandromontanari, line 102: Why are you logging only the first ten elements in `state` and `expected`? I suspect it is related to the bit logic you checked earlier with bitwise AND. Could you add a comment before this `log.warning` explaining why you are logging only the first ten elements?

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+            self.ref_state = state
+            self.ref_parity = parity
+            mismatches = np.where(state != expected)[0]
+            self.change[mismatches] += 1
+            self.event_times.append(trigger_time)
+            self.nchanges += 1
+            log.warning(
+                f"Reset reference. Changes incremented at indices: {mismatches[:10]}..."
```
alessandromontanari, line 113: If the reason why you are printing only the first 10 elements in `mismatches` is the same as for the `log.warning` just above, could you also add a comment above this line so the user knows it is all down to the same bit logic?

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+    def finish_run(self):
+        try:
+            self.change = np.array(self.change)
+            self.event_id = np.array(self.event_id)
+            self.event_times = np.array(self.event_times)
+
+            """self.run_start = (
+                self.event_times[self.event_id == np.min(self.event_id)] - 100
+            )
+            self.run_end = np.max(self.event_times) + 100"""
```
alessandromontanari: Do we need to keep these four commented lines? If not, could you remove them?

src/nectarchain/dqm/ping_pong.py
```diff
@@ -0,0 +1,165 @@
...
+                "CAMERA-PING-PONG-CHANGES-TIMES"
+            ] = self.event_times
+        except Exception as err:
+            log.error(f"Data could not be retrieved. Received error code: {err}")
+
+        return self.PingPongMonitoring_Results_Dict
+
+    def plot_results(self, name, fig_path):
+        try:
+            fig_pipo, disp = plt.subplots()
```
jlenain, line 144: `fig_pipo`, nice ! :laughing:

src/nectarchain/dqm/ping_pong.py
alessandromontanari: For the moment, I removed the commented `log.warning`. We can open an issue later if we need more verbose code for this processor

src/nectarchain/dqm/ping_pong.py
alessandromontanari: I added a comment explaining this check

src/nectarchain/dqm/ping_pong.py
alessandromontanari: Added a TODO for later, to explain in a comment why only logging the first ten elements.

alessandromontanari (approved):
Hi @jlenain, this PR is ready for merging. I added a very rudimentary unit test, which, in any case, will be improved later in my campaign to increase the test coverage anyway. Running this processor in the DQM locally correctly writes into the fits file:  
Image
And produces the plot for the ping pong changes:
Image

PR #337, opened 2026-06-22
Optimise performance DQM trigger statistics

alessandromontanari:
perf(src/nectarchain/dqm):
- trigger_statistics.py: several improvements in finish_run and get_results to create as few arrays as possible and return directly astropy Table
- dqm_summary_processor.py: include if clause for create_hdu for astropy Table
- start_dqm.py: optional tqdm, more efficient figures saving, logging times for several steps of the processor

This PR slightly optimises the performance of the `trigger_statistics.py` processor by saving a few seconds of computation. However, I believe this is not the processor from which we will save the most time, as there are not tight numerical loops that can be optimised with numba, not much more that can be vectorised or made faster with numpy arrays, and multiprocessing is not implemented yet. We'll probably optimise much more in the other, more computationally heavy, processors. Following this PR, there will be others optimising each processor.

### DIFF
--- a/src/nectarchain/dqm/dqm_summary_processor.py
+++ b/src/nectarchain/dqm/dqm_summary_processor.py
@@ -82,6 +82,11 @@ class DQMSummary:
 
             hdu = fits.BinTableHDU(data)
 
+        # Handle astropy Table objects directly
+        elif isinstance(content, Table):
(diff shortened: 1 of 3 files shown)


jlenain (approved):

PR #338, opened 2026-06-23
Optimise performance DQM waveforms

alessandromontanari:
This PR optimises significantly the `waveforms.py` processor in the DQM. 
Before optimisation, running only that single processor for both high and low gain, took around **94 s**. After optimisation, running with the same setup took around **31 s**! 
The PR includes a commit that I cherry-picked from #337. We shall first merge #337, and then I'll do a quick git rebase of this PR onto main, to avoid any conflicts for the merging.

### DIFF
--- a/src/nectarchain/dqm/dqm_summary_processor.py
+++ b/src/nectarchain/dqm/dqm_summary_processor.py
@@ -82,6 +82,11 @@ class DQMSummary:
 
             hdu = fits.BinTableHDU(data)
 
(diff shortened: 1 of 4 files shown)


jlenain (approved):
Thanks again, @alessandromontanari ! All good to me !

jlenain:
As far as I can see, the cherry-pick from #337 does not create any conflict with the `main` branch.
If that's fine with you, @alessandromontanari , we could merge this one right away ?

PR #340, opened 2026-06-24
Support several NectarCAM cameras in TRR verification package

### LINKED ISSUE #268: Support several cameras in TRR verification package
The different tests in the TRR verification package should support several NectarCAM cameras.

Each test should implement a `camera` argument, which would then need to be passed to the relevant classes, especially `DataManagement` when fetching raw data from CTADIRAC.

**Additional context**
At the moment, all the tests implemented in the TRR implicitly assume analyses of data acquired with the NectarCAM Qualification Model.

jlenain:
This PR addresses #268 by implementing the support of the different NectarCAM cameras (not only the QM one) in all the tests implemented in the TRR verification package.


src/nectarchain/trr_test_suite/charge_resolution.py
alessandromontanari, line 1: Hi @jlenain, I reviewed the changes on `charge_resolution.py`, and they are fine with me

src/nectarchain/trr_test_suite/charge_resolution.py
```diff
@@ -179,6 +192,7 @@ def main():
             pedestal_tool = PedestalNectarCAMCalibrationTool(
                 progress_bar=True,
                 run_number=run,
+                camera=camera,
```
alessandromontanari, line 195: I see that you pass `camera` as an argument to `PedestalNectarCAMCalibrationTool`, which however does not use it, does it? Did you have in mind implementing the `camera` argument later?

src/nectarchain/trr_test_suite/deadtime.py
alessandromontanari, line 1: I also reviewed these changes and I am fine with them.

src/nectarchain/trr_test_suite/deadtime.py
```diff
@@ -502,6 +514,7 @@ def run_deadtime_test_tool_process(
         tool = DeadtimeTestTool(
             progress_bar=True,
             run_number=run,
+            camera=camera,
```
alessandromontanari, line 517: Once again, the `camera` argument is not used by `DeadtimeTestTool`, but maybe it will be used later

alessandromontanari (commented):
Hi @jlenain, I reviewed all the changes and they look fine to me. 
I posted a couple of comments in the individual scripts, but I see that it is actually more general than I thought. Indeed, you pass the `camera` argument to several of the test tools, even though I think the argument is not used in the tools themselves. I think this is not code breaking, and in the end, it does not make much difference at the moment. But were you planning to implement something with this argument in the tools, maybe in a later PR?

src/nectarchain/trr_test_suite/charge_resolution.py
jlenain: These different tools inherit indirectly from the `BaseNectarCAMCalibrationTool` class, which uses the `camera` argument to load data from DIRAC through the `DataManagement.findrun` method in line 106 in `nectarchain.makers.core.py`. The functionality is thus already in place.

src/nectarchain/trr_test_suite/deadtime.py
jlenain: As above:

> These different tools inherit indirectly from the  `BaseNectarCAMCalibrationTool` class, which uses the `camera` argument to load data from DIRAC through the `DataManagement.findrun` method in line 106 in `nectarchain.makers.core.py`. The functionality is thus already in place.

jlenain:
> Hi @jlenain, I reviewed all the changes and they look fine to me. I posted a couple of comments in the individual scripts, but I see that it is actually more general than I thought. Indeed, you pass the `camera` argument to several of the test tools, even though I think the argument is not used in the tools themselves. I think this is not code breaking, and in the end, it does not make much difference at the moment. But were you planning to implement something with this argument in the tools, maybe in a later PR?

Hi @alessandromontanari !

Thanks a lot for the review ! Could you then please mark this PR as approved ?

To answer your comment about the `camera` argument in a general place:

These different tools inherit indirectly from the `BaseNectarCAMCalibrationTool` class, which uses the `camera` argument to load data from DIRAC through the `DataManagement.findrun` method in line 106 in `nectarchain.makers.core.py`. The functionality is thus already in place. The `camera` argument is passed along, implicitly to the base mother class.

alessandromontanari (approved):

PR #342, opened 2026-06-25
Fix unphysically large values in the waveforms DQM processor

### LINKED ISSUE #339: Inf or nan in the waveforms DQM processor
**Describe the bug**
The `waveforms.py` processor in the DQM writes to the fits file both arrays: the average waveforms and the waveforms for all pixels. Some pixels may be faulty for any unspecified run, and the waveform for those pixels may be set to inf or nan. This breaks the process of computing the average waveform in the `waveforms.py` processor. 

**To Reproduce**
1. from `/src/nectarchain/dqm`, `python start_dqm.py -c NectarCAMQM --noped --runnb 7213 --max-events 3000 $NECTARCAMDATA $NECTARCAMDATA`
2. explore the output fits file, and you will see that the entries for `WF-PHY-AVERAGE-PIX-HIGH-GAIN` or `WF-PHY-AVERAGE-PIX-LOW-GAIN` have values of ~2.56e+36 for each sample... 

**Expected behavior**
This is unphysical, and likely depends only on an untracked inf or nan in the pixel waveforms.

**Supporting information**
Image

**Additional context**
This is likely easy to fix by simply throwing an exception if the waveform for a specific pixel contains "unreasonable" values, and catching it by removing this pixel from the average?

jlenain:
This PR tries to fix #339 . I am not sure this is enough, but at least, reading a DQM FITS file back gives:

```python
from astropy.io import fits
hdr = fits.open("NectarCAM_Run7213_Results.fits")
hdr["WF-PHY-AVERAGE-PIX-LOW-GAIN"].data
FITS_rec([(inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,), (inf,),
          (inf,), (inf,), (inf,), (inf,)],
         dtype=(numpy.record, [('WF-PHY-AVERAGE-PIX-LOW-GAIN', '>f8')]))
```
which correctly flags these values as `inf` instead of unphysically large floats.

### DIFF
--- a/src/nectarchain/dqm/waveforms.py
+++ b/src/nectarchain/dqm/waveforms.py
@@ -108,12 +108,12 @@ class WaveFormsHighLowGain(DQMSummary):
         if self.counter_evt > 0:
             self.wf_average = self.wf / self.counter_evt  # get average
             # get average over pixels
-            self.wf_mean_over_pix = np.mean(self.wf_average, axis=0)
+            self.wf_mean_over_pix = np.nanmean(self.wf_average, axis=0)
 
         if self.counter_ped > 0:
             # get average pedestals
             self.wf_ped_average = self.wf_ped / self.counter_ped
-            self.wf_ped_mean_over_pix = np.mean(self.wf_ped_average, axis=0)
+            self.wf_ped_mean_over_pix = np.nanmean(self.wf_ped_average, axis=0)
 
         return None
 


jlenain:
I will close this PR, as it does not address #339 properly: using `np.mean` or `np.nanmean` does not change anything in the output DQM FITS files.

PR #349, opened 2026-07-02
Small "bug fix" to linearity range

pcorcam:
In the [FEBv6 performance paper](https://arxiv.org/abs/2311.11631), the linearity range between the two gain channels was found to be between 7–400 p.e. However, in more recent [thermal tests](https://redmine.cta-observatory.org/projects/nectarcam/wiki/NectarCAM_Thermal_Paper), the linearity range is quoted to be 3–200 p.e. After some discussion with @vmarandon, I decided to update the linear range to 10–200 p.e. in `nectarchain`, to be conservative. Note that in the end, this only affects the `HiLoNectarCAMCalibrationTool`.


jlenain (approved):
Thanks, @pcorcam !

PR #358, opened 2026-07-08
Display events trigger timestamps vs their ids on the DQM Bokeh app

alessandromontanari:
feat(src/nectarchain/dqm/bokeh_app):
- data/labels.json: add y-axis labels for trigger events timestamps
- app_hooks.py: add make_trigger_timestamps_vs_ids to make plots for timestamps vs event ids, and update_trigger_timestamps_vs_ids to work with the run selection update
- main.py: adapt layout with additional tab for the newly created plots
- tests/test_app_hooks.py: add tests
 
This PR addresses part of the additions to the Bokeh app which were discussed during the on-site coding sprint. In particular, it adds a tab with the trigger event timestamp vs id information, checking if the timestamps are correctly growing with time (so both the timestamps and the event ids are growing). If negative Timestamp[i+1] - Timestamp[i] are found, they are shown by overplotting red points on the blue ones. Moreover, the legend shows the count of how many negative timestamp differences are found.
We could envision adding a Div with another PR listing all the event timestamps and ids for which negative differences were found.

image

### DIFF
--- a/src/nectarchain/dqm/bokeh_app/data/labels.json
+++ b/src/nectarchain/dqm/bokeh_app/data/labels.json
@@ -11,6 +11,13 @@
         "CAMERA-BADPIXTIMELINE-PHY-HIGH-GAIN": "Fraction of bad pixels",
         "CAMERA-BADPIXTIMELINE-PHY-LOW-GAIN": "Fraction of bad pixels"
     }, 
+    "y_axis_labels_trig_timestamps": {
+        "TRIGGER-EVENTS-ALL": "Timestamps [MJD]",
+        "TRIGGER-EVENTS-PHY": "Timestamps [MJD]",
+        "TRIGGER-EVENTS-PED": "Timestamps [MJD]",
+        "TRIGGER-EVENTS-OTHERS": "Timestamps [MJD]",
+        "TRIGGER-EVENTS-WRONG": "Timestamps [MJD]"
+    },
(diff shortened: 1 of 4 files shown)


src/nectarchain/dqm/bokeh_app/data/labels.json
```diff
@@ -11,6 +11,13 @@
         "CAMERA-BADPIXTIMELINE-PHY-HIGH-GAIN": "Fraction of bad pixels",
         "CAMERA-BADPIXTIMELINE-PHY-LOW-GAIN": "Fraction of bad pixels"
     }, 
+    "y_axis_labels_trig_timestamps": {
+        "TRIGGER-EVENTS-ALL": "Timestamps [MJD]",
```
jlenain: From the example plot, the unit does not seem to correspond to MJD, but rather Unix time.

Also, in such a case, the number of digits is not sufficient.

alessandromontanari: You're right, I'll take care of it.

alessandromontanari:
Hi @jlenain, the last commit takes care of the UTC issue. The plots now look like below. I thought it would make more sense to show UTC time, and remove the date, as that is already displayed in the Div at the top of the app.

image

jlenain (approved):
Many thanks again, @alessandromontanari ! All good to me.

PR #362, opened 2026-07-23
Bokeh app categorise data for maker only once 

alessandromontanari:
perf(src/nectarchain/dqm/bokeh_app):
- extract_data.py: implement function to extract and categorise once and for all the data from the database, for the chosen run
- app_hooks.py: adapt maker and update functions to consume categorised data
- main.py: update main to input categorised data
- tests/test_app_hooks.py: update tests accordingly

This PR adds a function in `extract_data.py` to categorise the data for the chosen run only once. This avoids the repeated for loops in all the makers. `app_hooks.py` and `main.py` are refactored to consume the categorised data. Testing it locally, it saves around 0.10 s whenever the plots are made or updated, which is not much. However, it is still cleaner to have the categorisation done only once.

### DIFF
--- a/src/nectarchain/dqm/bokeh_app/app_hooks.py
+++ b/src/nectarchain/dqm/bokeh_app/app_hooks.py
@@ -2,7 +2,6 @@ import collections
 import json
 import os
 import re
-from datetime import datetime, timezone
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


jlenain (approved):
Thanks, @alessandromontanari, all good to me !

PR #363, opened 2026-07-23
DQM plots and output for peak time position avg and std

### LINKED ISSUE #345: Peak position in the DQM charge_integration processor
One of the outputs of `CtapipeExtractor.get_image_peak_time` in `src/nectarchain/dqm/charge_integration.py` is the peak time. This is extracted in line 148, and then appended to the list (depending on the event type) in lines 154, 160 or 168. However, this information is neither saved nor plotted later (cf. lines 172 and later).  

Shall we include it as an output, either by creating a plot for it or by saving it to the fits file? Is this information actually useful?

**Describe alternatives you've considered**
Either we think about how to use this output, or I may remove that list (and the process of appending new items to it) to save some memory.

**Additional context**
The last time the `peakpos` is used:
Image

alessandromontanari:
feat(src/nectarchain/dqm): 
- charge_integration.py: implement plots and output for peak time position avg and std

This PR implements part of the features requested by #345, by adding plots and output for `peakpos` and `peakpos_ped`, which were already extracted in the processor `charge_integration.py` but never added to the output. The output was tested locally with `python start_dqm.py -c NectarCAMQM --runnb 7213 --max-events 3000 --plot $NECTARCAMDATA $NECTARCAMDATA`.

### DIFF
--- a/src/nectarchain/dqm/charge_integration.py
+++ b/src/nectarchain/dqm/charge_integration.py
@@ -48,6 +48,8 @@ class ChargeIntegrationHighLowGain(DQMSummary):
         self.image_ped_stats = None
         self.ped_all_stats = None
         self.ped_ped_stats = None
+        self.peakpos_all_stats = None
+        self.peakpos_ped_stats = None
         self.ChargeInt_Results_Dict = {}
         self.ChargeInt_Figures_Dict = {}
         self.ChargeInt_Figures_Names_Dict = {}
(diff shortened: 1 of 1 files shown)


alessandromontanari:
Hi @jlenain, I was holding off this PR because of the discussion in #345, but since we've established that `2xmax-events`, depended on the slices not chained, I believe we could review and merge this. For now, it is adding plots for average and std of `peakpos`, in each pixel. The plots are created for high and low gain, and for `peakpos_ped` too, if it exists.

NectarCAM_Run7213_PeakPos_Mean_HighGain_All
NectarCAM_Run7213_PeakPos_Std_HighGain_All

The output are also written onto the fits files:

image

jlenain:

Do you think we could consider issue #345 fully solved once this PR is accepted and merged, or would you rather consider it closed once the information is also displayed in the Bokeh app ?

alessandromontanari:
> Thanks a lot, @alessandromontanari !
> 
> Do you think we could consider issue #345 fully solved once this PR is accepted and merged, or would you rather consider it closed once the information is also displayed in the Bokeh app ?

Hi @jlenain, yes I think so. This is at least one implementation of the feature request in #345. I will later implement the display in Bokeh, and if we need other quantities, we'll update the DQM processor in the future.

jlenain (approved):

jlenain:
Thanks a lot again, @alessandromontanari ! Everything looks good to me, a local test run also works like a charm.

PR #368, opened 2026-07-30
Small addition to correctly set up the `tel_id` in calibration files

pcorcam:
The `tel_id` is, by default, 0. Since the `PipelineNectarCAMCalibrationTool` does not require an event source directly - the subtools take care of that - it meant that in the cat-A calibration files, the `tel_id` was always set to 0. With this patch, we not only take care that the correct `tel_id` is used in the cat-A calibration files, but we also check that the different runs are from the same camera.


jlenain (approved):
Thank you, @pcorcam ! It is indeed good to check that all calibration coefficients are derived for the same camera. 

It all looks good to me.

PR #369, opened 2026-07-31
Clean Singularity/Apptainer cache at the end of DIRAC DQM jobs

jlenain:
Some grid sites warned me that our DQM jobs on DIRAC are letting some leftover in the local $HOME directory on worker nodes, most presumably coming from the Apptainer image cached locally when instantiating our containers there.

This PR adds a cleaning of the Apptainer cache before leaving production.


alessandromontanari (approved):
Hi @jlenain, I don't think I have of testing this myself. However, from your changes, I have understood that this should be the fix, run with $CALLER (singularity/apptainer) properly defined to clean the Apptainer/Singularity cache older than 3 days.

Therefore, I approve the PR

PR #371, opened 2026-08-04
DQM chain slices in one single EventSource

### LINKED ISSUE #366: DQM not chaining slices when creating EventSource
**Describe the bug**
As found out by @jlenain in #345, the `--max-events` argument is passed to an `EventSource`, for each and every run slice sitting on disk. It means that, e.g., for a run with 4 slices on disk, `--max-events=100` will actually result in an output with 400 events. This is somewhat due to a bad design in the DQM, where an `EventSource` is called for each run slice, instead of chaining slices in a single `EventSource`.

**To Reproduce**
1. Download all the slices for run 7213 
2. Run `python start_dqm.py -c NectarCAMQM --runnb 7213 --max-events 100 $NECTARCAMDATA $NECTARCAMDATA` on the full run (36 slices)
3. You'll get 3600 events

**Expected behavior**
There is no reason for not chaining slices. This can be corrected at the `start_dqm.py`, when creating the event reader with `EventSource`.

alessandromontanari:
fix(src/nectarchain/dqm): - start_dqm.py: fix loop creation of EventSource objects with the chaining of the input files in a single object instead

This PR fixes #366 by chaining the run data slices in a single EventSource instead of creating EventSource objects in loop for each slice.

### DIFF
--- a/src/nectarchain/dqm/start_dqm.py
+++ b/src/nectarchain/dqm/start_dqm.py
@@ -134,14 +134,8 @@ def main():
         log.error("Input files should be provided, exiting...")
         sys.exit(1)
 
-    # OTHERWISE READ THE RUNS FROM ARGS
-    path1 = args.input_files[0]
-
-    # THE PATH OF INPUT FILES
-    path = f"{NectarPath}/runs/{path1}"
-    log.debug(f"Input files:\n{path}")
-    for arg in args.input_files[1:]:
-        log.debug(arg)
+    all_paths = [f"{NectarPath}/runs/{arg}" for arg in args.input_files]
+    log.debug(f"Input files:\n{all_paths}")
 
     # Defining and printing the options
     PlotFig = args.plot
@@ -183,10 +177,6 @@ def main():
 
     start = time.time()
 
-    # INITIATE
-    path = path
-    log.debug(path)
-
     # Read and seek
     config = None
     if args.r0:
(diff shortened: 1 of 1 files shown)


alessandromontanari:
Hi @jlenain, in this PR I am changing `start_dqm.py` to use:
```
    reader = EventSource(
        input_filelist=all_paths, config=config, max_events=args.max_events
    )
    reader1 = EventSource(input_filelist=all_paths, config=config, max_events=1)
```
The main change is the usage of `input_filelist` argument, to receive a list of paths. In principle, I tested the file saving and plot production in local and it works for all processors except `ping_pong.py`--too many ticks in the color bar--and `pixel_timeline.py`--different x and y axis (however, I think these are unrelated errors that I will fix independently of using `input_filelist` instead of `input_url`). These two small problems are the reason why this is only a draft for now. 
Do you know if the way I am using `EventSource`, passing a list to `input_filelist`, is correct? I think so, from the results of the processors that work, but I wanted to check with you if I am missing something more fundamental from `ctapipe_io_nectarcam`.

jlenain:
> Hi @jlenain, in this PR I am changing `start_dqm.py` to use:
> 
> ```
>     reader = EventSource(
>         input_filelist=all_paths, config=config, max_events=args.max_events
>     )
>     reader1 = EventSource(input_filelist=all_paths, config=config, max_events=1)
> ```
> 
> The main change is the usage of `input_filelist` argument, to receive a list of paths. In principle, I tested the file saving and plot production in local and it works for all processors except `ping_pong.py`--too many ticks in the color bar--and `pixel_timeline.py`--different x and y axis (however, I think these are unrelated errors that I will fix independently of using `input_filelist` instead of `input_url`). These two small problems are the reason why this is only a draft for now. Do you know if the way I am using `EventSource`, passing a list to `input_filelist`, is correct? I think so, from the results of the processors that work, but I wanted to check with you if I am missing something more fundamental from `ctapipe_io_nectarcam`.

Hi @alessandromontanari ,

I think this is correct. In principle, one could also use a wildcard in `input_url`, but I am not sure it will work as expected. Under the hood, these arguments will trigger a call to the `ctapipe_io_nectarcam.MultiFiles` class (https://github.com/cta-observatory/ctapipe_io_nectarcam/blob/main/src/ctapipe_io_nectarcam/__init__.py#L566).

